PR #12, opened 2021-08-26
add register_bmi function to enable interoperablity with iso_c_bmif_2_0

hellkite500:
This PR will enable the noah-mp bmi module to be useable by the ngen framework via the ISO_C_Binding BMI module (noaa-owp/ngen#282) and the fortran BMI adapter.

This does introduce a dependency on the `iso_c_bmif_2_0` module, which can be found and built in the ngen subproject in `ngen/extern/iso_c_fortran_bmi`.  A PR on the ngen project with this library and the fortran adapter should be up shortly, will update this when it is.

### DIFF
--- a/modules/surface_bmi/bmi/Makefile
+++ b/modules/surface_bmi/bmi/Makefile
@@ -5,6 +5,8 @@
 
 include ../user_build_options
 
+F90FLAGS := -cpp ${F90FLAGS}
+
 OBJS = bmi_noahmp.o \
        bmi.o
 
--- a/modules/surface_bmi/bmi/bmi_noahmp.f90
+++ b/modules/surface_bmi/bmi/bmi_noahmp.f90
@@ -1,5 +1,7 @@
 module bminoahmp
 
+! NGEN_ACTIVE is to be set when running in the Nextgen framework
+! https://github.com/NOAA-OWP/ngen
 #ifdef NGEN_ACTIVE
    use bmif_2_0_iso
 #else
--- a/modules/surface_bmi/bmi/bmi_noahmp.f90
+++ b/modules/surface_bmi/bmi/bmi_noahmp.f90
@@ -1038,4 +1038,26 @@ contains
 !     call print_info(this%model)
 !   end subroutine print_model_info
 
+  function register_bmi(this) result(bmi_status) bind(C, name="register_bmi")
+   use, intrinsic:: iso_c_binding, only: c_ptr, c_loc, c_int
+   use iso_c_bmif_2_0
+   implicit none
+   type(c_ptr) :: this ! If not value, then from the C perspective `this` is a void**
+   integer(kind=c_int) :: bmi_status
+   !Create the momdel instance to use
+   !Definitely need to carefully undertand and document the semantics of the save attribute here
+   type(bmi_noahmp), target, save :: bmi_model !need to ensure scope/lifetime, use save attribute
+   !Create a simple pointer wrapper
+   type(box), pointer :: bmi_box
+
+   !allocate the pointer box
+   allocate(bmi_box)
+   !allocate(bmi_box%ptr, source=bmi_model)
+   !associate the wrapper pointer the created model instance
+   bmi_box%ptr => bmi_model
+   !Return the pointer to box
+   this = c_loc(bmi_box)
+   bmi_status = BMI_SUCCESS
+ end function register_bmi
+ 
 end module bminoahmp
--- a/modules/surface_bmi/bmi/bmi_noahmp.f90
+++ b/modules/surface_bmi/bmi/bmi_noahmp.f90
@@ -1,7 +1,12 @@
 module bminoahmp
 
+#ifdef NGEN_ACTIVE
+   use bmif_2_0_iso
+#else
+   use bmif_2_0
+#endif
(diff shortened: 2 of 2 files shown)


SnowHydrology (commented):
@hellkite500, In the current form, the surface module is no longer able to be compiled. Is there a way we can keep noah-mp-modular self-contained—i.e., it doesn't require ngen in order to run? A user could conceivably want to use the BMI-enabled version without running it through the ngen framework.

hellkite500:
I think there we can scheme up something that works.  The first two things that come to mind is implementing an independent module `ngen_module` which includes the noahmp_bmi and implements this function.  The other possibility is to use a preprocessor directive, and wrap the `register_bmi` function with `#ifdef NGEN_ACTIVE` or something like that.  The reason it doesn't compile is not finding the iso C module file.  So either of the approaches would give the ability for certain builds to not worry about it.

SnowHydrology:
Have you talked to the CSDMS folks about the [C-Fortran BMI extension](https://github.com/robertbartel/ngen/tree/f/fortran-bmi/main/extern/iso_c_fortran_bmi/src)? We could develop that as a standalone library that becomes a noah-mp-modular dependency _if_ running in the framework.

I hesitate to merge the PR because it means the bmi implementation of surface will no longer work. Do you want to chat tomorrow about this?

hellkite500:
So the only dependency of on the `iso_c_bmif_2.0` in the `register_bmi` function is due to a type in the module, `box` that is a trivial wrapper type.  So another option is to add the type definition of `box` to the module alongside `register_bmi` and then there would be no external dependency required.  But I'm not sure what that would do to the `iso_c_bmif_2.0` library that depends on that type.  It might work to duplicate it in both modules, but I'm not 100% sure about that at the moment.  Just wanted to save the idea here.

hellkite500:
@SnowHydrology There is no rush to merge this, I just wanted to get it up for this exact discussion.  I think we can definitely solve this problem one way or another to make it possible to compile/run the module completely independently.  Once we get that, then we can merge.

SnowHydrology:
@hellkite500 Sounds like a plan. Let me know when you want to move forward.

andywood:
I agree with Keith, I'd say it's a requirement that the noah-mp-modular
versions compile and run both in standalone and bmi-connected modes, and we
probably shouldn't merge until we have that. I like the compiler directive
solution (enabling a single driver), but alternatively having bmi & non-bmi
drivers would also work.  The driver doesn't do much, so it should not be
hard to keep two relatively synced.  But either way, we should be able to
solve this.

Cheers,
-Andy

andywood:
It may be better process to make a new branch for this in the OWP repo, then redirect this fork PR to that branch before finally merging it with master.  it may make it easier to reverse course (hopefully not needed but ...) and document the development changes, merge etc.

hellkite500:
@SnowHydrology @andywood I implemented the CPP preprocessor option, so this should allow the code to compile stand alone (as long as `NGEN_ACTIVE` isn't defined to the preprocessor.  Seems to work well for now.

SnowHydrology:
@hellkite500, It looks like we still need to add the new BMI specification to the `/bmi` directory and that information to the makefile.

hellkite500:
> @hellkite500, It looks like we still need to add the new BMI specification to the `/bmi` directory and that information to the makefile.

I don't understand, this should work entirely independent of the "new BMI specification".  I thought that was the intention.

SnowHydrology:
> > @hellkite500, It looks like we still need to add the new BMI specification to the `/bmi` directory and that information to the makefile.
> 
> I don't understand, this should work entirely independent of the "new BMI specification". I thought that was the intention.

Probably just my confusion, but if someone sets NGEN_ACTIVE to true, then won't the module need access to `bmif_2_0_iso`? Or is `NGEN_ACTIVE` something we will only define when using the framework? If it's the latter, I'll just add a comment and approve the PR. Thanks!

hellkite500:
> if someone sets NGEN_ACTIVE to true, then won't the module need access to `bmif_2_0_iso`?

Yes, but the expectation is that the NGEN build system will set this flag and make the `bmif_2_0_iso` module for building the NGEN shared library.  For any "core" dependency, I see NGEN being responsible for compiling/linking external modules, such as noah-mp.  [Here](https://github.com/NOAA-OWP/ngen/blob/master/extern/noah-mp-modular/CMakeLists.txt) is what we do for the current noah-mp module, this CMake file builds the dependent iso module, sets the NGEN_ACTIVE cpp directive, and then builds the fortran sources for noah-mp and the surfacebmi module.  In this case noah-mp doesn't need to worry about the framework integration other than the few lines of code wrapped up in the cpp directives.  

It is possible to replicate this behavior directly in the dependent modules.  If you want to be able to build a NGEN compatible shared library from your project's own build files, then you have to manage that iso dependency, which requires a more sophisticated build processes (auto conf, cmake, ninja, ect.) I don't think we want to get in the habit of copying source code files around too many places, but instead should try to manage them as real dependencies.

SnowHydrology:
That's a good explanation and makes sense to me. I added two comment lines above the directive. Are you good with a merge?

hellkite500:
I'm good, thanks for the comment commit!

PR #13, opened 2021-09-07
move model parameters from namelist.input

nmizukami:
- Split namelist (commit e7d2af6b7fb16c375bdb9fec3f58728e0df1a963)
  previous - `namelist.input` includes model configurations, initialization and parameters 
  update - `param.namelist` includes model parameters and `namelist.input` includes model configurations, initialization.

- Split namelist reading routine (commit 41adfcc342f134b24a510d6055b73eff7026060d)
  Based on the namelist split, the namelist reading routine needs to be split into two - ParametersRead.f90 and 
  modelConfigRead.f90

- Other changes based on new namelist routines (commit eee05fb39f7f45a2c4df73dedb281a2397751624 and ccf44cc59808e7ba75816bc7483140df41889c50)
  Various type.f90 files and Makefiles


andywood (commented):
Hi Naoki, I think this is a good first step, and separates parameters from other model options & settings.  Ideally, though, we want to get back to reading the standard noah-mp TABLES (eg VEGPARM.TBL, GENPARM.TBL, MPTABLE.TBL etc), so that we can easily leverage upgrades in parameters by the noah-mp community.  So instead of a namelist parameter read, we implement the something resembling HRLDAS parameter read.  Having two namelists for the code may be a bit confusing.

andywood:
I can merge this now if it runs, so that the PR doesn't get too big.

nmizukami:
Yes, I remember this conversation (using *.TBL). I will look into this. I tested this and get identical result with the original one (just used nccmp)

SnowHydrology:
> I can merge this now if it runs, so that the PR doesn't get too big.

@andywood, please don't merge as I'd like to take a deeper look tomorrow at some of the things that stood out to me. Thanks for getting this started, @nmizukami!

andywood:
Hi Keith,

Sorry, by 'now' I meant 'after Keith looks it over too'.  I mainly meant
that I think this gets us halfway where we want to be -- but if it's
usable, I can merge it so that the next part doesn't turn this into a
massive PR.  Also, quick note, the merge is from Naoki's fork branch into
the OWP branch on param reads, not into master.

-Andy

modules/full/driver/noahmp_driver.f90
```diff
@@ -8,7 +8,8 @@ program noahmp_driver
   use NoahMPOutput
   use LevelsType
   use DomainType
-  use NamelistRead
+  use modelConfigRead
```
SnowHydrology, line 11: Is there a strong motivation to change from `NamelistRead` to `modelConfigRead`? I noticed in the code below that we still use the `namelist_type`, so this may introduce unnecessary confusion. If we choose the latter name, I would prefer it just say `ConfigRead`.

modules/full/driver/noahmp_driver.f90
```diff
@@ -50,12 +52,14 @@ program noahmp_driver
...
   logical            :: precipitating     ! .true. if precipitating
   real               :: QV_CURR           ! water vapor mixing ratio (kg/kg)
+  character(len=256) :: param_nml_file = '../parameters/param.namelist'
 
 !---------------------------------------------------------------------
 !  initialize
 !---------------------------------------------------------------------
 
   call namelist%ReadNamelist()
+  call paramNML%ReadNamelist(param_nml_file)
```
SnowHydrology, line 62: These two calls use different syntax for the same operation. `namelist%ReadNamelist()` gets the file name from `modelConfigRead`, whereas `paramNML%ReadNamelist(param_nml_file)` gets the file name from the driver. I have little preference, but I'd like them to be consistent.

modules/full/driver/noahmp_driver.f90
```diff
@@ -26,7 +27,8 @@ program noahmp_driver
 !  types
 !---------------------------------------------------------------------
 
-  type (namelist_type)     :: namelist
+  type (modelConfig_type)  :: namelist
+  type (parametersNML)     :: paramNML
```
SnowHydrology, line 31: Does NML here stand for namelist? It seems like we're introducing unnecessary uncertainty by calling the params file a namelist and having two parameters types.

modules/full/parameters/param.namelist
```diff
@@ -0,0 +1,161 @@
+&soil_parameters
```
SnowHydrology, line 1: I agree with @andywood that we should not call this a namelist and that the params should be split out into their respective tables if we want to go from leaving everything in the namelist as we currently do to using a method that's more consistent with the HRLDAS version of Noah-MP.

modules/full/run/namelist.input
```diff
@@ -65,164 +65,6 @@
...
- ISBARREN                  = 16
- ISICE                     = 15
- ISCROP                    = 12
- EBLFOREST                 =  2
- NATURAL                   = 14
- LOW_DENSITY_RESIDENTIAL   = 31
- HIGH_DENSITY_RESIDENTIAL  = 32
- HIGH_INTENSITY_INDUSTRIAL = 33
+&parameters 
+ parameters_nml = ./parameters/param.namelist
```
SnowHydrology, line 69: This is also in the driver. Does it serve an additional purpose here?

modules/full/src/DomainType.f90
```diff
@@ -40,7 +40,7 @@ module DomainType
   subroutine Init(this, namelist)
 
     class(domain_type) :: this
-    type(namelist_type) :: namelist
+    type(modelConfig_type) :: namelist
```
SnowHydrology, line 43: To echo my earlier comment, I don't think introducing a new type (`modelConfig_type`) just to rename it to `namelist` is the way to go here.

SnowHydrology (changes requested):
Let's address some of the comments before merging. Thanks for the great work, @nmizukami!

modules/full/driver/noahmp_driver.f90
andywood: I'd call it ParamRead, since the configuration is largely set by the namelist.  Again, avoid confusion since other models' control files are referred to as config.  This is a parameter reading routine.

modules/full/driver/noahmp_driver.f90
andywood: agree.  there should be no confusion between read namelist (the config) and read params.

modules/full/run/namelist.input
andywood: I think of the driver assignments should eventually get moved to other source files, keeping the driver clean.

andywood:
Sounds good, Naoki.  I also don't think the namelist should contain initial
state values -- that would be part of a readState function & file set.
Ideally the namelist would include:
* major filenames / paths
* time control settings (start/stop/timestep)
* space control settings (eg subdomain)
* physics options
* maybe output options (eg frequency)

nmizukami:
@SnowHydrology and @andywood please take a look.   I cleaned up (remove parameterRead.f90) and put paramRead under parameter type data structure.

nmizukami:
I accidentally removed tailing spaces (my vim editor setting for fortran code and Makefile etc.)

andywood:
merging this PR for now.  there may be small things left we can clean up but let's keep those commits for the next PR.

PR #15, opened 2021-09-13
noah-mp parameter tables

nmizukami:
put noahmp parameter tables (reading routing not implemented yet)


SnowHydrology:
@nmizukami, is this PR for the new tables only or will you be adding more?

nmizukami:
I believe those two (GENPAR and SOIL) only. MPTABLE (also needed) is already there.

SnowHydrology:
Thanks, @nmizukami, I'll merge these tables in.

PR #18, opened 2021-09-16
Dev unix time

andywood:
Adding unix time functionality; times now included in output files.  Tested on Bondville example.

### DIFF
--- a/modules/full/src/DateTimeUtilsModule.f90
+++ b/modules/full/src/DateTimeUtilsModule.f90
@@ -961,6 +961,7 @@ contains
   
   subroutine get_utime_list (start_datetime, end_datetime, dt, times)
     ! makes a list of data times in secs since 1970-1-1 corresponding to requested period
+    ! reports end-of-timestep points
     implicit none
  
     real*8, intent (in)               :: start_datetime, end_datetime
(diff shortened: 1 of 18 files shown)


nmizukami (commented):
I tried running with this update and looks good to me. One comment on style - there are some old pre-fortran90 style statement left (one example is double precision statement). I thought it would be good to update?

modules/full/src/DateTimeUtilsModule.f90
```diff
@@ -0,0 +1,998 @@
...
+    g = mod (e, p) / r
+    h = u * g + w
+    day = mod (h, s) / u + 1
+    month = mod (h/s+m, n) + 1
+    year = e / p - y + (n+m-month) / n
+  
+  end subroutine calendar_date
+ 
+ 
+  double precision function date_to_unix (date)
```
nmizukami, line 877: since kr8 is defined for double precision maybe use real(kr8)? double precision statement is pre-fortran 90 style and is obsolete now.

andywood:
Since these are minor comments that don't affect code performance, and I'm short on time, I'm going to merge this and we can come back to it to add the tweaks in a subsequent PR.

PR #19, opened 2021-09-17
Replacing param.namelist (parameter file) with noahmp TBLs

nmizukami:
These changes enable to use the parameters from noahmp TBLs (GENPARM.TBL  MPTABLE.TBL, SOILPARM.TB).

NOTE: 
1. Z0 is hard-coded in parameter_type%paramRead subroutine (L276). This is also hardcoded in ENERGY subroutine in HRLDAS version noahmp.
2. SHDFAC is inserted in MPTABLE.TBL for now.  This used to be in VEGPARM.TBL. VEGPARM.TBL no longer exists in HRLDAS. Need to figure out where this is located. Looks like wrfhydro still uses VEGPARM.TBL (and older MPTABLE.TBL)


andywood (commented):
I think it's preferable not to hardwire the MPTABLE name.  If different versions of tables exist, one may want to switch them easily without having to link or copy the versions into that exact name.  This is why I took the hardwired approach out of summa and put the table names into the config file, eg

! --- input table file names ---
vegTableFile        'VEGPARM.TBL'                    ! vegetation param table
soilTableFile       'SOILPARM.TBL'                   ! soil param table
generalTableFile    'GENPARM.TBL'                    ! general param tables and misc params
noahmpTableFile     'MPTABLE.TBL'                    ! noah-mp related param tables

Incidentally, we could do something similar to SUMMA and create a 'GENPARM.TBL' that has misc parameters we may want to add (such as the Z0), rather than stick them in MPTABLE. eg, 

General Parameters
SLOPE_DATA
9
0.1 
0.6
1.0
0.35
0.55
0.8
0.63
0.0
0.0
SBETA_DATA
-2.0
FXEXP_DATA
2.0
CSOIL_DATA
2.00E+6
SALP_DATA
2.6
REFDK_DATA
2.0E-6
REFKDT_DATA
3.0
FRZK_DATA
0.15
ZBOT_DATA
-8.0
CZIL_DATA
0.1
SMLOW_DATA
0.5
SMHIGH_DATA
3.0
LVCOEF_DATA
0.5

nmizukami:
I put noahmp table names in configuration namelist (namelist.input) and also put Z0 in GENPAR.TBL.

andywood (commented):
This looks to be moving in the right direction.  I'm going to merge for now and we should review to see what more we want to do on parameter reads.

PR #20, opened 2021-09-20
Ensure ReadNamelist reads from the passed parameter file

hellkite500:
Fixes a bug where the BMI initialization code incorrectly tries to read part of the namelist input from a hardcoded filename `namelist.input` instead of the passed `config_file`.


SnowHydrology (approved):
Thanks @hellkite500

PR #21, opened 2021-09-20
Allow noah-mp ReadNamelist to read from optional filename arg

hellkite500:
Proposed change to the core, or "full", noah-mp module to allow the namelist reading to configured from a passed argument.  The argument is optional, so it shouldn't break any existing calls that do not provide it.  This is related to #20, but I didn't want to include it there if the `bmi` and `full` modules were intentionally heavily decoupled.  If this change were pushed to the full module, then it may be that there would be no need to maintain two different `NamelistRead.f90` modules and two different `ReadNamelist` subroutines.

### DIFF
--- a/modules/full/src/NamelistRead.f90
+++ b/modules/full/src/NamelistRead.f90
@@ -198,10 +198,15 @@ end type namelist_type
      
 contains   
 
-  subroutine ReadNamelist(this)
+  subroutine ReadNamelist(this, namelist_file)
   
     class(namelist_type) :: this
-    
+    ! Optional namelist_file path/filename to read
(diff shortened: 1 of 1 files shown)


andywood:
I thought we had this functionality already built into the 'surface' version, but looking at that right now I still see the hardwired name.  I agree, we definitely want to have the config file name be an argument.

hellkite500:
It is in the surface version, #20 fixes a small bug there.  Just wasn't sure if it was worth putting this into the "full" version.

andywood (commented):
It makes sense to reconcile the versions.  I'm going ahead to merge.

PR #31, opened 2021-10-08
preprocessor directives for NGEN_FORCING and NGEN_OUTPUT

### LINKED ISSUE #22: Add Nextgen flags to turn off forcing data read and output write functionalities
When running in the Nextgen model engine, we should have flags that allow forcing data and output functionalities to be ignored so that IO is handled exclusively by Nextgen. There is currently an NGEN_ACTIVE flag set for the BMI specification, so we should add granular flags that refer to forcing and output.

SnowHydrology:
This PR adds two preprocessor directives to `modules/surface_bmi`:

1. `NGEN_FORCING_ACTIVE`: When the surface module is run in Nextgen, we can now turn off the module's internal handling of forcing data. When `NGEN_FORCING_ACTIVE` is defined (`#define NGEN_FORCING_ACTIVE`), Nextgen **must** supply forcing data via BMI or some other means.
2. `NGEN_OUTPUT_ACTIVE`: Similar to the above, we can now let Nextgen write (or not write) output data instead of using the surface module's internal routine. When `NGEN_OUTPUT_ACTIVE` is defined (`#define NGEN_OUTPUT_ACTIVE`), Nextgen **should** write output data.

Closes #22

### DIFF
--- a/modules/surface_bmi/driver/Makefile
+++ b/modules/surface_bmi/driver/Makefile
@@ -5,6 +5,8 @@
 
 include ../user_build_options
 
+F90FLAGS := -cpp ${F90FLAGS}
+
 OBJS = noahmp_output.o noahmp_driver.o noahmp_ascii_read.o
 
 all:	$(OBJS)
--- a/modules/surface_bmi/driver/noahmp_output.f90
+++ b/modules/surface_bmi/driver/noahmp_output.f90
@@ -1,5 +1,10 @@
 module NoahMPOutput
 
+!---------------------------------------------------------------------
+! Compiler directive NGEN_OUTPUT_ACTIVE to be defined if 
+! Nextgen is writing model output (https://github.com/NOAA-OWP/ngen)
+!---------------------------------------------------------------------
+#ifndef NGEN_OUTPUT_ACTIVE
   use netcdf
   use WaterType
   use EnergyType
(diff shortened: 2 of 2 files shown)


nmizukami:
> > Logics looks good me (I don't have any comments in codes). The other way around, if you run this module as stand-alone (ie. w/o ngen), NGEN_FORCING_ACTIVE is not defined so use module's internal routine (hopefully this is correct?).
> 
> I'm not a preprocessor expert, but I believe that is correct. The surface module should run as-is (reads its own forcing, writes its own output) unless the flags are defined in the build somewhere.

I am not really expert on this either, but I also believe this is correct.  I think we will need to test with ngen.

SnowHydrology:
Okay, I'm going to make some edits to ensure NetCDF can be ignored if `NGEN_OUTPUT_ACTIVE` is defined. One thing I noticed, which may be obvious, is that the surface module runs _much_ faster when not writing to NetCDF.

SnowHydrology:
I've amended the output file so that it's an empty module when `NGEN_OUTPUT_ACTIVE` is defined. This should mean we're good to go now.

If, for some reason, someone would want to run the surface module without writing output they should change their `user_build_options`:

- Add `-DNGEN_OUTPUT_ACTIVE`
- Change ` NETCDFLIB      =       -L/usr/local/Cellar/netcdf/4.8.0_2/lib -lnetcdf -lnetcdff` to ` NETCDFLIB      =       #-L/usr/local/Cellar/netcdf/4.8.0_2/lib -lnetcdf -lnetcdff` so that NETCDFLIB is an empty var

PR #32, opened 2021-10-25
Reading noahmp table for parameter initialization in surface-bmi

nmizukami:
Added three noah-mp tables: SOIL, MPTABLE (updated to the recent HRLDAS), and GENPARM, and 
a module (ParameterRead.f90) including noah-mp table reading routine

NOTE: This commit just added the above files, and table reading routine has not been tied to the rest of code yet. so the code just work as before.


andywood:
Since **surface_bmi** version is being used by the framework team, we may not want to do any new merges to that version to avoid affecting their runs.

nmizukami:
this is obsolete (and invalid).  so closed

PR #33, opened 2021-11-03
BMI implmentation in modules/full

SnowHydrology:
This PR moves the BMI implementation into the full module and includes all associated changes to ensure BMI compliance. 

## Additions

- `bmi` subdirectory that includes BMI module and functions
- `src/RunModule.f90` to execute most driver-level functionalities executed via BMI
- `data` subdirectory to store input/output

## Changes

- Makefile edits to ensure correct dependencies

## Todos

- Add evaporation fix from `surface_bmi` module
- Add runoff option set for coupled and internal


andywood (commented):
I think we don't want this inclusive setting because it includes one more timestep than is in the run, with different variables available for the start or end timestep depending on the convention.  For instance, if the convention is that the datetime stamp of an output record is start of period, then the very last record will be missing variable data.  If it is for end of period, then the very first datetime stamp will be missing data.  It's conventional that the output of a model includes the # of records in the run.  I changed this earlier when adding the time/date info.   Worth discussing?

andywood:
you're right.  I just tried to clarify in slack that it assumes the start & end times given denote the full time period covered by the run, but if we have a different convention for that then we do need to add back the +1 as you suggest.  in any case, good catch to revisit this.

SnowHydrology:
I added the NGEN flags for forcing and output.

PR #34, opened 2021-11-08
Aligning modular version with monolith codebase

SnowHydrology:


SnowHydrology:
I've made some small changes to ensure the code is aligned through the end of the `EnergyModule` with the `noahmp_sflx` subroutine and its associated calls in the monolith Noah-MP codebase.

SnowHydrology:
I've now implemented `opt_sub` so that `modules/full` can reproduce either Noah-MP subsurface behavior (`opt_sub == 1`) or mimic the surface module with a hydrostatic subsurface (`opt_sub == 2`).

andywood (commented):
Looks good and kudos for adding good documentation & code/comment alignment.  One suggestion in EnergyModule is to trap unexpected option values in the if(option = x) constructions by adding an 'else' clause, but that can be added later.

PR #36, opened 2021-11-23
Ending support for multiple modules, source code to main level

### LINKED ISSUE #17: Refactor code and organizational structure
## Current behavior

Noah-MP-Modular is split into three separate modules: `full`, `surface`, and `surface_bmi`. If code is modified in one module, then the corresponding code has to be modified in the others.

## Desired behavior

Implement the hydrostatic subsurface as an option in `full` to reduce the amount of edits that need to be made when code is modified.

SnowHydrology:
This PR eliminates the triplicate codebase of the original noah-mp-modular. No longer will we support `modules/full`, `modules/surface`, and `modules/surface_bmi`. Instead we will have a single codebase that runs using BMI with an option for a full Noah-MP-style subsurface or a simplified subsurface for use when coupling to a runoff module.

Closes #17


andywood (approved):
go for it

SnowHydrology:
PR now ready for review.

Major changes:

- The former `full` module is the only module supported now
- To operate it in surface mode, set `subsurface_option = 2` in `namelist.input`
- Source code and other subdirectories moved from the former `modules/full` path to the main level of the noah-mp-modular directory
- The full module runs by default with BMI, there is no longer a supported non-BMI model version
- BMI output variables have been added for model evaluation when running in Nextgen
  - `TG` (ground surface temperature, K) and `SNEQV` (snow water equivalent, mm)
- Due to the impending name change, the executable is now `noah_owp_modular.exe` 
- Parameter tables are now in the `parameters` subdirectory instead of `namelist.input`
  - Although this is not a change for the `full` module, it is a change from how `modules/surface_bmi` is currently running in Nextgen

Minor change:

- Added OWP GitHub documentation

PR #38, opened 2021-12-02
Compiler directive to fix precipitation issues when running in Nextgen

SnowHydrology:
Noah-OWP-Modular now works in both standalone and Nextgen modes.


SnowHydrology:
Poor PR with old commit history

PR #39, opened 2021-12-02
Compiler directive to fix precip issues when running in Nextgen

SnowHydrology:
Noah-OWP-Modular now works in both standalone and Nextgen modes.

### DIFF
--- a/src/AtmProcessing.f90
+++ b/src/AtmProcessing.f90
@@ -63,8 +63,13 @@ contains
 
     ! sum different precip types to get total
     !PRCP = forcing%PRCPCONV + forcing%PRCPNONC + forcing%PRCPSHCV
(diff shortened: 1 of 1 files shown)


andywood:
looks good.  maybe not quite a bug - more that the framework wasn't full updated to match the latest model code so the code adaptation lets it work until we make the modes consistent.  anyway, since PRCP is now in the forcing data type and gets updated this should work just fine for our standalone output too.

If we do stick with the convention that the prcpvar going into the model is categorized as 'nonc', then we may want to add back those other terms that were originally there (in case it's later coupled to WRF, etc.).  For now they're zero and so they won't make a difference:      !orig:  PRCP = forcing%PRCPCONV + forcing%PRCPNONC + forcing%PRCPSHCV

PR #43, opened 2022-01-13
repo sweep of old model references

SnowHydrology:
This cleans up references to the Noah-MP land surface model except where we are specifically referring to NCAR's version of the model.


andywood:
I don't actually know -- I'll ask.  It could be 'model parameter' instead
of 'multi-physics', for instance.  In any case, I lean toward renaming the
table file (ie keeping your PR as is), as long as we make clear in the repo
documentation that MPTABLE.TBL is the template for NOAHOWP.TBL.

SnowHydrology:
That works for me. Because we could technically pass any file name to noahowp_table via the namelist, this should be good for now. Let me know if you hear anything on what the MP stands for in MPTABLE.

PR #48, opened 2022-04-14
Adjust incoming solar radiation for slope and aspect

### LINKED ISSUE #47: Correct incoming solar radiation for slope and aspect
Noah-OWP-Modular assumes its model domain is flat. This means it cannot resolve the effects of slope and aspect on hydrologic processes modulated by incoming solar radiation (e.g. snowmelt, evapotranspiration). This should be rectified.

We will need to modify the following files to incorporate the effects of slope and aspect on solar radiation. The main change will be to produce a new cosine of the solar zenith angle (`COSZ`) value.

-`src/UtilitiesModule.f90` (modify equations for `COSZ`)
-`src/NamelistRead.f90` (read in slope/aspect)
-`src/DomainType.f90` (add slope/aspect)
-`run/namelist.input` (add slope/aspect)

SnowHydrology:
This PR updates Noah-OWP-Modular to account for slope and aspect when computing incoming solar radiation. Previously the model assumed a flat domain, meaning it could not resolve the effects of topography on various hydrologic processes.

Files changed:

-`src/UtilitiesModule.f90` (modified equations for `COSZ`)
-`src/NamelistRead.f90` (read in slope/aspect)
-`src/DomainType.f90` (add slope/aspect)
-`run/namelist.input` (add slope/aspect)

If you want to run Noah-OWP-Modular assuming flat topography, you can just set `terrain_slope` and `azimuth` in `namelist.input` to 0.0.

Closes #47

### DIFF
--- a/run/namelist.input
+++ b/run/namelist.input
@@ -18,6 +18,8 @@
 &location ! for point runs, needs to be modified for gridded
   lat              = 40.01    ! latitude [degrees]
   lon              = -88.37  ! longitude [degrees]
+  terrain_slope    = 0.0      ! terrain slope [degrees]
+  azimuth          = 0.0      ! terrain azimuth or aspect [degrees clockwise from north]
 /
 
 &forcing
--- a/src/DomainType.f90
+++ b/src/DomainType.f90
@@ -25,6 +25,8 @@ type, public :: domain_type
   real                :: lat               ! latitude (°)
   real                :: lon               ! longitude (°)
   real                :: ZREF              ! measurement height of wind speed (m)
+  real                :: terrain_slope     ! terrain slope (°)
+  real                :: azimuth           ! terrain azimuth or aspect (° clockwise from north)
   integer             :: vegtyp            ! land cover type
   integer             :: croptype          ! crop type
   integer             :: isltyp            ! soil type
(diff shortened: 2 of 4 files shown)


andywood (commented):
Great addition of functionality.  Aside from the physics/code, one comment related to usage is whether the /location/ block is now more of an /attributes/ or even potentially /local_parameters/ collection of information, and would be better moved out of the namelist.   I tend to like a trim namelist that holds runtime info, some broad settings that have to be read first (eg structure dimensions for using in allocating arrays).  Noah-om doesn't have any input files right now that are for location specific attributes or parameters, but it may be good to start one or both -- it would be used in calibration.  Genparm & the other tables are the kinds of things you might use across an entire domain, and wouldn't want to require individually for each catchment.  if the application expands to run vectors or grids, these would have to go into another file (versus arrays in the namelist).  not to have mentionitis around SUMMA, but it has one input file for geo-attributes (which don't change, such as lat/lon/slope/aspect, and one for local calibratable parameters, which is the only one updated in a calibration loop.  I kind of like the arrangement.   *just a thought*

andywood:
@keith, I can go through the math or some tests of the algorithm on the solar rad. adjustment, but I wonder if you've also done some sensitivity testing to see if the coded implementation gives expected results (eg n/s/e/w facing slopes at different locations)?

SnowHydrology:
image

I used the equations from Dewalle and Rango (2008), which are not as extensive as the protocol implemented in SUMMA. One concern is that once lat_eq reaches a maximum of 90°, it starts to go back towards the equator for steep slopes with north-facing azimuths at higher latitudes (55°N shown). This may prove moot based on declination. What are your thoughts?

SnowHydrology:
It looks like I should also compute corrected sunrise and sunset times for the slope/azimuth, and limit the computation of incoming solar radiation to hours in between. Noah-OM limits solar calcs to periods when COSZ > 0, so this should be a straightforward fix.

SnowHydrology:
Changed method for slope-azimuth correction from equivalent latitude and hour angle adjustment to the vector solution from Corripio (2003): https://doi.org/10.1080/713811744

SnowHydrology:
Merged the PR into snow mods after running some simple tests (changing slope/azimuth and examining output). We might want to identify more rigorous testing for edge cases.

PR #49, opened 2022-04-25
Implementing get_ptr_* functions to return BMI_FAILURE

zhengtaocui:
TYPE: Feature enhancements

KEYWORDS: BMI 2.0, get_value_ptr_*

SOURCE: Zhengtao Cui, NOAA NWS OWP

DESCRIPTION OF CHANGES:

1.
As we discussed in the Fortran and BMI meeting (Keith, Scott, Matt, Nels and Zhengtao), we concluded that

bmi.f90 changes to enable get_value_ptr means that updated code would not be compatible with other Fortran models
We likely will not use get_value_ptr, so changing the spec to use the function would not be beneficial (and we could degrade performance actually per Donald)
Given these issues, we will NOT implement the changes required to use get_value_ptr. It will be set to return BMI_FAILURE (see point below).
Few Fortran models use target attributes and pointers, so it may be useful to escalate this issue moving forwards (Scott to note at BMI roundtable)
Update our Fortran spec, and return BMI_FAILURE for non-implemented functions (Nels and I will coordinate)
Scott has worked on BMI extensions in C and Fortran
BMI roundtable is happening soon, at which changes will be proposed and accepted/rejected
Trey has indicated he wants the other extensions ready to go
We can keep the conversation going as time progresses (keep model engine in the loop with BMI roundtable decisions and C++/Python extension progress)
ME team should consider going over extension to make sure there are no major concerns or, alternatively, important suggestions

Because we won't be modifying get_value_ptr to return values, we should update the BMI implementation in Noah-OWP-Modular to change the commented out functions to return BMI_FAILURE. Although not ideal, this is technically in line with BMI protocol as it allows the functions to display expected behavior.

2.
Added a build option for Linux with a gfortran compiler, such as Hera.

ISSUE:  

TESTS CONDUCTED: A test code similar to the existing NoahModularDriver.f90 in driver/ directory is created to test these get_value_ptr_* functions. The test code is not checked in, however.

NOTES: 1) Tested only with the gfortran compiler. Need to test with an Intel compiler. 2) The test code for get_value_ptr_functions is not checked into the repository.


Makefile
```diff
@@ -13,4 +13,14 @@ clean:
...
+	(cd run;		make clean)
+
+testBMI:
+	(cd src;		make)
+	(cd bmi;		make)
+	(cd test;		make)
+
+testBMI_clean:
+	(cd src;		make clean)
+	(cd bmi;		make clean)
```
SnowHydrology, line 25: Between the `bmi` and `test` lines, add: `	(cd driver;		make clean)`

zhengtaocui:
I just wanted to merge the Makefile changes to my repo. I don't know if I did it the right way or not. It looks like I've merged it to NOAA-OWP repo instead. If this is the case, please let me know. I am not familiar with github.

SnowHydrology:
@zhengtaocui, you merged the commits from the original PR, but my comments are unresolved. This is okay because I changed the destination branch to `bmi_update` from `master`. Please do the following:

1. Do a fresh clone of the repo
2. Fetch all branches
3. Checkout the `bmi_update` branch locally
4. Make a new branch from `bmi_update` (call it `fix_makefile` or something descriptive)
5. Make the changes I noted in my comments
6. Stage and commit them
7. Push the changes to GitHub
8. Open a new pull request and tag me as a reviewer

Please let me know if you need any help with this. Thanks!

zhengtaocui:
Keith,

I made a new branch 'fix_makefiles' of the 'bmi_updates' branch and created
a pull request. Please review.

Thanks for the detailed instructions.

Zhengtao

zhengtaocui:
Keith,

Thanks for the information. For steps 1 to 7, should I do them in my
terminal or in the browser of the github site?

Zhengtao

PR #57, opened 2022-07-15
Merge snow-mods branch to main

### LINKED ISSUE #46: Add precipitation phase option(s) that incorporate humidity
Noah-OWP-Modular currently relies on air temperature alone to determine precipitation phase. These methods will induce inaccuracies in modeled SWE. See [Jennings et al. (2018)](https://www.nature.com/articles/s41467-018-03629-7) and [Jennings and Molotch (2019)](https://hess.copernicus.org/articles/23/3765/2019/) for more info.

We should considering adding one or two additional methods:

1. A wet bulb temperature threshold
2. A binary logistic regression where phase is predicted as f(air temperature, relative humidity)

Additionally, the air temperature thresholds are hard-coded. We should consider moving them to the namelist.

To incorporate these changes, we will need to modify:

- `src/AtmProcessing.f90`
- `src/NamelistRead.f90` (if adding threshold as a parameter)
- `src/ParametersType.f90` (if adding threshold as a parameter)
- `src/OptionsType.f90` (comment documentation)
- `run/namelist.input`

### LINKED ISSUE #47: Correct incoming solar radiation for slope and aspect

### LINKED ISSUE #53: Add new EnergyType variable to approximate ground surface temperature
The `TG` variable in Noah-OWP-Modular is nominally ground temperature, but it looks like it becomes snow surface temperature when there is snow on the ground. This means we no longer have information on ground surface temperature that we can pass to a subsurface module such as Soil Freeze Thaw.

To do: Make a new scalar variable that is equal to `TG` when there is no snow on the ground and equal to the lowest snow element temperature `STC[0]` when there is snow on the ground.

Need to change:

- `src/EnergyType.f90` (new var that takes the `TG` or `STC[0]` value)
- `src/EnergyModule.f90` (assigns value to new var in if-else statement)
- `bmi/bmi_noahowp.f90` (specifies var info for passing via BMI)

SnowHydrology:
This PR will merge the `snow-mods` branch into main. It includes a selection of improvements to the Noah-OWP-Modular codebase designed to better represent cold regions hydrology. These are:

1. Two additional precipitation phase partitioning methods to split precipitation into rain and snow. One is a wet bulb temperature threshold and the other is a binary logistic regression model that estimates the probability of snowfall as a function of air temperature and relative humidity. Merged to `snow-mods` in PR #51 (click for more info).
2. Routines to correct incoming solar radiation for slope and aspect. The original Noah-MP assumed a flat domain which is appropriate running at large spatial scales, but less so at the smaller scales of the National Water Model. Merged to `snow-mods` in PR #52 (click for more info).
3. A new ground surface temperature variable for passing to the Soil Freeze Thaw model. The original code switches `TG` to be snow surface temperature when snow is on the ground. A new routine adds `TGS` which is equivalent to `TG` when there is no snow or the temperature of the top soil layer when there is snow. Merged to `snow-mods` in PR #56 (click for more info).

I've tested these modifications in both standalone mode and in the Next Generation Water Resources Modeling Framework.

Closes #46
Closes #47 
Closes #53

### DIFF
--- a/src/AtmProcessing.f90
+++ b/src/AtmProcessing.f90
@@ -30,6 +30,10 @@ contains
     real, parameter         :: RHO_GRPL = 500.0  ! graupel bulk density [kg/m3] ! MB/AN : v3.7
     real, parameter         :: RHO_HAIL = 917.0  ! hail bulk density [kg/m3]    ! MB/AN : v3.7
     real                    :: QV_CURR           ! water vapor mixing ratio (kg/kg)
+    real                    :: rs_thresh         ! local var for rain-snow threshold (takes user-defined or hardcoded value, depending on option)
+    real                    :: temp              ! temperature (can be air or wet bulb depending on opt_snf option)
+    real                    :: rh                ! relative humidity (computed for opt_snf 6 and 7)
+    
     ! --------------------------------------------------------------------------------------------------
 
     ! Compute derived variables from forcing data
@@ -47,6 +51,7 @@ contains
     energy%TAH = forcing%SFCTMP                         ! assign canopy temp with forcing air temp (K)
     QV_CURR    = forcing%Q2 / (1 - forcing%Q2)          ! mixing ratio, assuming input forcing Q2 is specific hum.
     energy%EAH = forcing%SFCPRS*QV_CURR/(0.622+QV_CURR) ! Initial guess only. (Pa)
+    
 
     ! Set incoming shortwave to 0 if cosine of solar zenith angle < 0
     IF(energy%COSZ <= 0.0) THEN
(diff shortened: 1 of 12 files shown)


src/AtmProcessing.f90
```diff
@@ -92,50 +110,83 @@ SUBROUTINE ATM (options, parameters, forcing, energy, water)
...
+        snow_prob = 1/(1 + exp(-10.04 + 1.41 * tair_C + 0.09 * rh)) ! compute probability as f(tair, rh)
+        
+        ! Compute FPICE as function of probability
+        IF(snow_prob >= 0.5) THEN
+          forcing%FPICE = 1.0
+        ELSE
+          forcing%FPICE = 0.0
+        ENDIF
+      
+      case default
```
nmizukami, line 187: I am not sure if there exist a error handling routine in NOAH-owp (if there is run time error, terminate the program and tell what is wrong), it would be good to add it here. if options%OPT_SNF is somehow more than 7, tell the user that options%OPT_SNF is invalid.

src/EnergyModule.f90
```diff
@@ -348,6 +348,15 @@ SUBROUTINE EnergyMain (domain, levels, options, parameters, forcing, energy, wat
     
     ! derived diagnostic variables
     energy%LH = energy%FCEV + energy%FGEV + energy%FCTR
+    
+    ! Compute TGS, or ground surface temperature for passing to a subsurface module
+    ! TGS is equal to TG when SNOWH <= 0.05 and equal to the temperature of the bottom snow
+    ! element when SNOWH > 0.05
+    IF (water%SNOWH <= 0.05) THEN
```
nmizukami, line 355: try not use <=, >=, == for real number comparison?  so could write like this.
```
IF (water%SNOWH > 0.05) THEN
     energy%TGS = energy%STC(0)
ELSE
     energy%TGS = energy%TG
END IF
```

nmizukami (commented):
these are some picky comments.

src/AtmProcessing.f90
SnowHydrology: @nmizukami I think we should add this for all option flags when they are read in. I'll open an issue and start a new branch.

src/EnergyModule.f90
SnowHydrology: Modified with commit

SnowHydrology:
@nmizukami, I updated the real number comparison and will perform the error handling for all option flags in a new PR. This PR can be merged when you're ready. Thanks!

src/AtmProcessing.f90
nmizukami: Yes, maybe create error handling subroutine that elegantly terminate the program with a message (so you could pass some meaningful message).  See if i can find good example. 

I think yes, you can leave this as for this PR and implement error handling maybe throughout the program in a separate PR

nmizukami:
Hi @SnowHydrology , I don't have any further comments so if you and Andy are ok. please go ahead and merge.

andywood:
Hi Keith and Naoki,
Just to comment briefly on the above discussion:
* I agree with Keith that these models are lacking input error checking of
all kinds, and we should seek to add this where we think users are most
likely to make errors, or where the lack of a trap is going to lead to
mysterious behavior.   This is a good example of where it's a good idea;
ultimately it's subjective how far to go with that.   In my experience, the
more the model gets used, the more such handing will end up being
added, and no models have it everywhere they could.
* On the real number comparisons, is this a concern about machine error
issues representing #s, eg 0.3?   If so, we could consider defining the
tolerances we want to use and including them in the comparisons
explicitly.  Otherwise this is a new one to me so I'm curious about it.

SnowHydrology:
@andywood and @nmizukami, I added the error handling as issue #58

PR #60, opened 2022-08-05
Error handling routine

### LINKED ISSUE #58: Include error handling for out-of-range options
There is currently no internal error handling for out-of-range options in Noah-OWP-Modular. This means we can enter a non-sensical value for an option flag and the model would still try to run without alerting us of an issue.

We should implement code, likely in `NamelistRead.f90` that checks each option to make sure the provided value is within a valid range. If the option is not in range, then the model should stop running and return a clear message. In pseudo code:

```
if option_val < 0 or option_val > option_max then
    print "option_val for option_name out of range, exiting..."
    stop
end if
```

nmizukami:
- Added new module including error abort routine. 

- One use case is to check model option check. 

resolve #58

### DIFF
--- a/src/NamelistRead.f90
+++ b/src/NamelistRead.f90
@@ -1,5 +1,7 @@
 module NamelistRead
 
+use ErrorCheckModule, only: sys_abort
+
 implicit none
 save
 private
@@ -83,7 +85,7 @@ contains
     ! Temporary var to hold the default, "namelist.input"
     ! or the value of namelist_file, if passed
     character(:), allocatable :: namelist_file_
-    
+
     integer            :: iz
     real               :: dt
     character(len=12)  :: startdate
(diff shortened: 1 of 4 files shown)


nmizukami:
thanks Andy,   I think I saw utilitiesModule, but this mainly contains date and time related routines (and solar geometry), so I did not put this there.

zhengtaocui:
I agree with Andy that this module can be generalized and could be further enhanced to be used in a broader scope.

I see there is another option called 'structure_option' that has a value of either 1 or 2. It is not in the 'model_option' group but in the 'structure' group. Do you think we should add range checking for this variable as well?

nmizukami:
thank you @zhengtaocui, i will put check on structure_option. Also if you have any suggestion on new module (now ErrorCheckingMod.f90)- module name, any modification on subroutine in the module, any other routines (will take a look at Andy's netcdf routine)?

src/NamelistRead.f90
```diff
@@ -187,8 +189,31 @@ subroutine ReadNamelist(this, namelist_file)
...
+    if (frozen_soil_option<1 .or. frozen_soil_option>2) then; call sys_abort(1,'model options: frozen_soil_option should be 1-2'); end if
+    if (dynamic_vic_option<1 .or. dynamic_vic_option>3) then; call sys_abort(1,'model options: dynamic_vic_option should be 1-3'); end if
+    if (dynamic_veg_option<1 .or. dynamic_veg_option>9) then; call sys_abort(1,'model options: dynamic_veg_option should be 1-9'); end if
+    if (snow_albedo_option<1 .or. snow_albedo_option>2) then; call sys_abort(1,'model options: snow_albedo_option should be 1-2'); end if
+    if (radiative_transfer_option<1 .or. radiative_transfer_option>3) then; call sys_abort(1,'model options: radiative_transfer_option should be 1-3'); end if
+    if (sfc_drag_coeff_option<1 .or. sfc_drag_coeff_option>2) then; call sys_abort(1,'model options: sfc_drag_coeff_option should be 1-3'); end if
+    if (canopy_stom_resist_option<1 .or. canopy_stom_resist_option>2) then; call sys_abort(1,'model options: sfc_drag_coeff_option should be 1-2'); end if
+    if (crop_model_option/=0) then; call sys_abort(1,'model options: crop_model_option should be 0'); end if
```
zhengtaocui, line 206: Maybe we can get rid of this option if only 0 is allowed.

zhengtaocui (commented):
How about we add a is_within_bound function to the ErrorCheckModule?

function is_within_bound (lower_bound, upper_bound, var) result( withinbound )
     implict none
     integer, intent(in) :: lower_bound
     integer, intent(in):: upper_bound
     integer, intent(in):: var
     logical  :: withinbound

     if ( var < lower_bound .or. var > upper_bound ) then
        withinbound = false
     else
         withinbound = true
end function

src/NamelistRead.f90
nmizukami: crop_model_option is removed

zhengtaocui:
Thank you Naoki. I am merging this enhancement to the base.

PR #64, opened 2022-08-11
Develop

andywood:
This PR adds error checking to the namelist file read and makes a few other minor changes.  The new namelist is shorter and has a few different labels:  'input_file' is changed to 'forcing_file', one group is removed and one has a name changed.  

The error checking finds namelist entries that are not recognized and also missing required namelist entries.  It introduces integerMissing, realMissing, and stringMissing variables that can be used more broadly in the code if needed. 

The code was tested against the original develop branch and an NCDIFF on the output showed no differences.   The unit test was run and passed.  The new functionality was tested for missing int, real and string entries. 

Lastly, this includes a small fix after it was noticed that the code to handling a run where the user does not supply a namelist did not work.


zhengtaocui:
@andywood It looks good. I just want to make sure that by removing the structure_option, there is no optional keys in the namelists, all keys are mandatory.

andywood:
It looks like with 'crop_type_option' removed, 'crop_type' could also be removed, and also the Namelist in the run directory should be updated to remove both.   I just resolved conflicts from the develop branch that had my namelist checking, but this was at the same time as some other activity so they may not have stuck.  I'll wait until other activity settles and look at it again.

zhengtaocui:
Andy, I just resolved some conflicts with the base branch and updated the unit testing files. I've committed my updates. You can go ahead and commit your changes now.

andywood:
Thanks Zhengtao.
I think we've decided that all keys are mandatory.  I haven't checked to
see if there's any key that might never be checked in a model run, but if
there were one, we could change to allow that to be missing.
We do have at least one option that isn't strictly necessary (crop_option)
-- it's left in for now because we might want to add the crop code back in
at some point.

The structure_option seemed unnecessary because it just set up a special
case that can be easily handled by the user using normal entries.
But I wanted to make sure that it had not been added because it was needed
by the framework in some way.

Cheers,
Andy

PR #71, opened 2023-06-21
homogenous gridded driver for noah-owp-modular

GreyEvenson-NOAA:
# PURPOSE

This PR is meant accomplish the first of three steps/milestones aimed at creating a gridded version of NOAA-NWS OWP's Noah-mp model (i.e., 'Noah-owp'). Step 1 is:

1. Develop Noah-owp capacity to execute over a homogenous gridded domain

Steps 2 and 3 will be accomplished via subsequent PRs. Steps 2 and 3 are:

2. Develop Noah-owp capacity to execute over a heterogenous gridded domain
3. Develop Noah-owp capacity to execute using real-word spatial data

# ADDITIONS

This PR adds a series of new 'grid' derived data types, each defined in a distinct module, i.e.:

  - `domaingrid_type` in DomainGridType.f90
  - `energygrid_type` in EnergyGridType.f90
  - `forcinggrid_type` in ForcingGridType.f90
  - `levelsgrid_type` in LevelsGridType.f90
  - `optionsgrid_type` in OptionsGridType.f90
  - `parametersgrid_type` in ParametersGridType.f90
  - `watergrid_type` in WaterGridType.f90

The PR-added 'gridded' derived data types correspond the original derived data types within 
the Noah-owp column model, i.e.:

  - `domaingrid_type` corresponds to `domain_type`, as defined in DomainType.f90
  - `energygrid_type` corresponds to `energy_type`, as defined in EnergyType.f90
  - `forcinggrid_type` corresponds to `forcing_type`, as defined in ForcingType.f90
  - `levelsgrid_type` corresponds to `levels_type`, as defined in LevelsType.f90
  - `optionsgrid_type` corresponds to `options_type`, as defined in OptionsType.f90
  - `parametersgrid_type` corresponds to `parameters_type`, as defined in ParametersType.f90
  - `watergrid_type` corresponds to `water_type`, as defined in WaterType.f90

The PR-added 'gridded' derived data types contain the same variable names that exist within their corresponding
original derived data types (e.g., both `water_type` and `watergrid_type` have a variable named `qinsur`); however, the 'gridded' version of the variables (e.g., `qinsur` in `watergrid_type`) have x and y dimensional indices if the variable varies over those dimensions. For example, `qinsur` is defined as `real,allocatable,dimension(:,:) :: qinsur` in `watergrid_type` but is defined as `real :: qinsur` in `water_type`.

[...]

PR #74, opened 2023-08-01
Add functionality to read gridded domain info and land use from input netcdf file

GreyEvenson-NOAA:
# PURPOSE

This PR modifies the gridded version of the Noah-OM model to read gridded domain info (e.g., `n_x`, `n_y`, `dx`, and `dy`) and land use (i.e., `vegtyp`) values from a NetCDF input file. Subsequent PRs will add functionality to read additional gridded model variables (e.g., soils, slope, etc).

# ADDITIONS

This PR adds `GridInfoType.f90`, which defines `gridinfo_type`. The `gridinfo_type` has member variables to hold read-in gridded input variables. The `gridinfo_type` has one `public` type-bound subroutine (`gridinfo_type%ReadGridInfo`).

The PR-added `public` `gridinfo_type%ReadGridInfo` subroutine is primarily intended to make calls to the `gridinfo_type`'s private type-bound subroutines.

The PR-added subroutine `gridinfo_type%ReadSpatialInfo` reads spatial information (i.e., `n_x`, `n_y`, `dx`, `dy`, and latitude and longitude values) from the NetCDF file identified by its arguments. The subroutine (i.e., `gridinfo_type%ReadSpatialInfo`) is intended to be called for every NetCDF file (with the assumption that every gridded input variable will be provided via a separate NetCDF file) to ensure that all variables are provided over the same grid.

This PR adds the 'private' type-bound subroutine `gridinfo_type%ReadVegtyp`, which is called within `gridinfo_type%ReadGridInfo`. The primary purpose of `gridinfo_type%ReadVegtyp` is to read-in gridded land use data. 

This PR adds `data/netcdf_dummy.nc` to facilitate testing. It is a dummy NetCDF file having gridded land use data in the format expected by this PR's code additions. The latitude, longitude, and `vegtyp` values for grid cell (1,1) match those for the control model run as listed in the pre-PR `namelist.input `file such that the output from grid cell (1,1) should match the control run while the output from other grid cells is expected to vary.

This PR adds `test/analysis/create_dummy_netcdf.py` was added to create `data/netcdf_dummy.nc`. This script was added to facilitate changes to the organization and contents of the input NetCDF file per PR feedback.

# CHANGES

`namelist.input` was modified to remove the scalar `vegtyp` entry and add an entry for `vegtyp_filename`, which is the directory/name of the input NetCDF file passed to `gridinfo_type%ReadVegtyp`.

Changes were made to several of the initialization subroutines for the gridded derived data type members of `noahowpgrid_type`. Several of these subroutines were modified to take an instance of the `gridinfo_type` as an argument, often to provide access to `n_x` and `n_y` for allocating allocatable arrays. Note that land use values are transferred from `gridinfo_type` to `domaingrid_type` when `gridinfo_type` is given as an argument to `domaingrid_type%InitTransfer`.

# TESTING

The PR-modified model was successfully compiled and executed in stand-alone on MacOS using GNU Fortran (Homebrew GCC 13.1.0) 13.1.0. 

The PR-added example input NetCDF file (i.e., `data/netcdf_dummy.nc` contains gridded land use data in the format expected by this PR's code additions. The latitude, longitude, and `vegtyp` values for grid cell (1,1) match those for the control model run as listed in `run/namelist.input` in the main branch. Thus, model output from grid cell (1,1) should match the control run while the output from other grid cells may vary. `test/analysis/compare_outputnc.py` was executed and I observed that all tests passed for grid cell (1,1) while at least one or more tests failed for all other grid cells -- this was the expected result. The output from `test/analysis/compare_outputnc.py` is: [output.txt](https://github.com/NOAA-OWP/noah-owp-modular/files/12294926/output.txt)

The unit tests (`test/noahowp_driver_test.f90`) were re-executed and output was normal. The output from `test/noahowp_driver_test.f90` is: [output.txt](https://github.com/NOAA-OWP/noah-owp-modular/files/12294936/output.txt)

# NOTES

I'm expecting one gridded model input per NetCDF file. In this case, we can add additional private type-bound subroutines to open and read from specific NetCDF files. Each added type-bound subroutine must make a call to `gridinfo_type%ReadSpatialInfo` immediately after opening the NetCDF file to ensure that the variable is given on the expected grid.

However, if we elect to assume that all gridded inputs will be provided in a singular NetCDF, this code can be easily modified to facilitate that expectation. 

# Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/)


src/Makefile
GreyEvenson-NOAA, line 1: This file was updated to facilitate this PR (#74) *and* PR #71. I'm not sure why PR #71 worked without updates to this file but it did.

data/netcdf_dummy.nc
SnowHydrology, line 1: See suggested changes in my review comments.

run/namelist.input
SnowHydrology, line 1: Is the plan to have all gridded values be in one NetCDF file or spread across multiple. If the latter, then you'll want to make the file name here more illustrative.

SnowHydrology (changes requested):
@GreyEvenson-NOAA, thanks for getting this started. A few changes:

- Can you modify the `netcdf_dummy.nc` to have valid latitudes and longitudes, consistent with the provided `dx` and `dy` values?
- Can you also vary the land cover values within the file so that we can verify that the land cover values do indeed vary within the grid and that we get differing output based on land cover?
- In the changed `netcdf_dummy.nc` file, there should be one grid cell that matches the parameters of the test case (vegtyp, lat, lon, etc.), while the others differ per my recommended changes above.
- I'm not sure how I feel about the name `gridlist`. It's not a namelist, so there's no reason to keep the "list" part. Maybe change it to `gridinfo` or something like that.

run/namelist.input
GreyEvenson-NOAA: I'm going back and forth with this question. On one hand, I can see why having a single NetCDF file would be preferable as it will allow for NetCDF-defined dimensions to be shared across multiple NetCDF-defined variables. This could cut down on problems attributable to a model-user providing variables that do not share the same grid (?). On the other hand, we could expect a separate NetCDF file for every variable and write checks to stop the program if there is any difference in grid info across the provided grids. Having a separate NetCDF file and associated line in namelist.input for separate model variables also just seems more intuitive (word choice?) to me.

Do you have thoughts/preferences regarding this choice, @SnowHydrology ?

GreyEvenson-NOAA: I got feedback from Scott on a standup advocating for using a single NetCDF file for each of the model's gridded input variables (e.g., one for land use, a second for soils, etc. etc.) but got feedback from Nels on slack advocating use of a single NetCDF for all gridded inputs. I can imagine scenarios in which a model-user would prefer either and so I don't have a firm opinion either way but revised the code expecting multiple files. 

Note that I believe the code should still work if the model-user simply puts all gridded variables in one file and gives the path/name to a single NetCDF input file on multiple lines of `namelist.input` (e.g., on separate lines for land use, soils, slops, etc.). This would lead to some sunk costs in opening/closing the file multiple times and reading dimensions multiple times but it's an option.

GreyEvenson-NOAA:
@SnowHydrology, I think we're ready for you to have another look. Note that I revised the PR-description to incorporate your comments. But to address the specific points in your previous comment:

- `netcdf_dummy.nc` now has valid lat/longs in agreement with provided dx and dy values following revisions to the script used in its creation (`test/analysis/create_dummy_netcdf.py`)
- The revised  `netcdf_dummy.nc` has the lat, long, and land use values from the test case (i.e., `run/namelist.input` in main branch) while all other grid cells have different lat, long, and land use values.
- I changed `gridlist_type` to `gridinfo_type`

PR #80, opened 2023-08-17
Add capacity to read-in gridded soils data from NetCDF

GreyEvenson-NOAA:
# PURPOSE

This PR adds capacity to read-in gridded soils data from an input NetCDF file. The path/name of the NetCDF file must be provided by the model-user in `namelist.input`.

# ADDITIONS

Added the `private` type-bound subroutine `gridinfo_type%ReadSoils`. The `ReadSoils` subroutine takes one argument (`filename`), which is the path/name of the NetCDF input file containing soils data as specified by the model-user in `namelist.input`. The purpose of this subroutine is to read soils data from the NetCDF input file to populate/set `gridinfo_type%isltyp`, which will be transferred to `domaingrid_type%isltyp` with a subsequent call to `domaingrid_type%InitTransfer`.

`data/netcdf_soils.nc` was added as an example soil NetCDF file and to facilitate testing. This file has the same spatial information/dimensions as `data/netcdf_landuse.nc` and has a soil type of 1 for grid cell (1,1), which is consistent with the soil type specified in `run/namelist.input` in this repo's main branch corresponding to the unmodified column model. All other grid cells have randomly assigned soil types. 

# CHANGES

The subroutines `gridinfo_type%ReadLandUse` (renamed from `gridinfo_type%ReadVegtyp`) and `gridinfo_type%ReadSpatial` were revised such that the NetCDF files are opened and closed within the scope of only those subroutines. My thought was that this approach would better satisfy our desire for 'separation of concerns' though there is an extra computation cost of opening/closing a NetCDF file more than what is absolutely necessary. 

The public type bound subroutine `gridinfo_type%ReadGridInfo` was revised to include additional actions (i.e., allocating the `gridinfo_type` arrays and calling `ReadSpatial`).

`data/netcdf_landuse.nc` was renamed from `data/netcdf_vegtyp`.

# TESTING

The PR-modified model was successfully compiled and executed in stand-alone on MacOS using GNU Fortran (Homebrew GCC 13.1.0) 13.1.0.

The test script (i.e., `test/analysis/compare_outputnc.py` which compares gridded model output against output from the unmodified column model) was executed. As the testing input NetCDF files depicts grid cell (1,1) as having a soil type (`isltyp`) of 1 and a land use (`vegtyp`) of 1 -- which is consistent with `run/namelist.input` in this repo's main branch -- the expected behavior of the script is that grid cell (1,1) will pass all tests while all other grid cells should fail at least one test. The output from `test/analysis/compare_outputnc.py` is: [output.txt](https://github.com/NOAA-OWP/noah-owp-modular/files/12381568/output.txt)

The unit test program was re-executed and its output was observed normal. The output is: [output.txt](https://github.com/NOAA-OWP/noah-owp-modular/files/12381572/output.txt)

# NOTES

The `gridinfo_type%ReadLandUse` and `gridinfo_type%ReadSoils` subroutines are nearly identical. Hence, I considered removing these subroutines and replacing them with a single generalizable subroutine (e.g., something like `gridinfo_type%ReadVariable`). However, it seems like we'd need to create a separate `ReadVariable` subroutine for each variable data type (i.e., `integer`, `real`) so we'd end up having, for example, `ReadVariable_int` and `ReadVariable_real`. I'm fine going that way but I am currently expecting to simply create a single subroutine for each variable that will be read into the model (soils, land use, slope, etc). The adopted approach seemed like it might afford more flexibility if variable-specific logic checks need to be executed (e.g., check that land use values are within expected bounds).


src/GridInfoType.f90
```diff
@@ -266,19 +361,27 @@ subroutine ReadSpatial(this,filename,ncid)
...
+      if(read_nx /= this%n_x) then; write(*,*) 'ERROR : x-dimension resolution from ''',trim(filename),''' does not match x-dimension resolution from other gridded inputs'; stop; end if
+      if(read_ny /= this%n_y) then; write(*,*) 'ERROR : y-dimension resolution from ''',trim(filename),''' does not match y-dimension resolution from other gridded inputs'; stop; end if
+      if(read_dx /= this%dx)  then; write(*,*) 'ERROR : x-dimension spacing from ''', trim(filename),''' does not match x-dimension spacing from other gridded inputs';  stop; end if
+      if(read_dy /= this%dy)  then; write(*,*) 'ERROR : y-dimension spacing from ''', trim(filename),''' does not match y-dimension spacing from other gridded inputs';  stop; end if
       do ix = 1, read_nx
-        if (all(read_lon(ix) /= this%lon(ix,:))) then; write(*,*) 'ERROR : lat values from ''', trim(filename),''' do not match lat values from other gridded inputs';  stop; end if
+        if (all(read_lon(ix) /= this%lon(ix,:))) then; write(*,*) 'ERROR : longitude values from ''', trim(filename),''' do not match longitude values from other gridded inputs';  stop; end if
```
nmizukami, line 369: Supposed to be `any` instead of `all`?   The same for latitude

src/GridInfoType.f90
```diff
@@ -246,10 +341,10 @@ subroutine ReadSpatial(this,filename,ncid)
...
-      if(read_ny /= integerMissing) then; this%n_y = read_ny; else; write(*,*) 'ERROR : problem reading n_y from ''',trim(filename),''''; stop; end if
-      if(read_dx /= realMissing)    then; this%dx  = read_dx; else; write(*,*) 'ERROR : problem reading dx from ''', trim(filename),''''; stop; end if
-      if(read_dy /= realMissing)    then; this%dy  = read_dy; else; write(*,*) 'ERROR : problem reading dy from ''', trim(filename),''''; stop; end if
+      if(read_nx /= integerMissing) then; this%n_x = read_nx; else; write(*,*) 'ERROR : problem reading x-dimension resolution from ''',trim(filename),''''; stop; end if
+      if(read_ny /= integerMissing) then; this%n_y = read_ny; else; write(*,*) 'ERROR : problem reading y-dimension resolution from ''',trim(filename),''''; stop; end if
+      if(read_dx /= realMissing)    then; this%dx  = read_dx; else; write(*,*) 'ERROR : problem reading x-dimension spacing from ''', trim(filename),''''; stop; end if
+      if(read_dy /= realMissing)    then; this%dy  = read_dy; else; write(*,*) 'ERROR : problem reading y-dimension spacing from ''', trim(filename),''''; stop; end if
       if(read_lon(1) /= realMissing) then
```
nmizukami, line 348: `this%lat` and `this%lon` are 2D arrays (type component), while `read_lat` and `read_lon` are 1D array (local array).  Here copying lat array vertically in `this%lat` and and copying lon array horizontally in `this%lon`. This works for rectilinear grid (regular lat/lon grid), but not for curvilinear grid.  Actually curvilinear grid always comes in 2D arrays.   I wonder if you deal with only rectilinear grid?   if so, why `this%lat` and `this%lon` needs to be 2D??

src/GridInfoType.f90
```diff
@@ -266,19 +361,27 @@ subroutine ReadSpatial(this,filename,ncid)
...
       do ix = 1, read_nx
-        if (all(read_lon(ix) /= this%lon(ix,:))) then; write(*,*) 'ERROR : lat values from ''', trim(filename),''' do not match lat values from other gridded inputs';  stop; end if
+        if (all(read_lon(ix) /= this%lon(ix,:))) then; write(*,*) 'ERROR : longitude values from ''', trim(filename),''' do not match longitude values from other gridded inputs';  stop; end if
       end do
       do iy = 1, read_ny
-        if (all(read_lat(iy) /= this%lon(:,iy))) then; write(*,*) 'ERROR : lat values from ''', trim(filename),''' do not match lat values from other gridded inputs';  stop; end if
+        if (all(read_lat(iy) /= this%lat(:,iy))) then; write(*,*) 'ERROR : latitude values from ''', trim(filename),''' do not match latitude values from other gridded inputs';  stop; end if
       end do
 
     end if
```
nmizukami, line 375: I see `else` corresponding to this `end if` easily, but not `if xxx then` (I am sure it is somewhere...)

nmizukami (commented):
Hi @GreyEvenson-NOAA,  overall looks great, but I have some question on what type of grid can be handled - rectilinear (just regular lat lon grid) and curvilinear (it coming from projected coordinate).  you can find the comment where you populate `this%lat` and `this%lon`.

test/analysis/create_dummy_netcdf.py
```diff
@@ -1,51 +1,102 @@
...
+  print(var_lat)
+  print(var_isltyp)
+
+  # Close dataset
+  dum.close()
+
+def create_dummy_vegtyp():
+
+  # Create and open dataset
+  filename = os.path.join(loc,'netcdf_landuse.nc')
```
nmizukami, line 60: yes, I think this is good way to construct file path (windows use backslash and old specification does not work)

src/GridInfoType.f90
```diff
@@ -87,73 +107,136 @@ subroutine ReadVegtyp(this,filename)
     if (.not. lexist) then
       write(*,*) 'ERROR Could not find ''',trim(filename),''''; stop ":  ERROR EXIT"
     endif
```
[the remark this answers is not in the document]
GreyEvenson-NOAA: True. Changed here and for all code responsible for opening the NetCDF files throughout this module.

src/GridInfoType.f90
```diff
@@ -87,73 +107,136 @@ subroutine ReadVegtyp(this,filename)
...
+              var            => this%vegtyp, &
+              integerMissing => this%integerMissing)
+
+    !----------------------------------------------------------------------------
+    ! Check that file exists and open it
+    !----------------------------------------------------------------------------
+    inquire(file = trim(filename), exist = lexist)
+    if (.not. lexist) then
+      write(*,*) 'ERROR Could not find ''',trim(filename),''''; stop ":  ERROR EXIT"
+    endif
```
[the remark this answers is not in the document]
GreyEvenson-NOAA: True. Changed here and for all code responsible for opening the NetCDF files throughout this module.

src/GridInfoType.f90
```diff
@@ -185,32 +268,44 @@ subroutine ReadSpatial(this,filename,ncid)
     dimid_x = integerMissing
     dimid_y = integerMissing
 
+    !----------------------------------------------------------------------------
+    ! Check that file exists and open it
+    !----------------------------------------------------------------------------
+    inquire(file = trim(filename), exist = lexist)
+    if (.not. lexist) then
+      write(*,*) 'ERROR Could not find ''',trim(filename),''''; stop ":  ERROR EXIT"
+    endif
```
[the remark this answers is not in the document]
GreyEvenson-NOAA: True. Changed here and for all code responsible for opening the NetCDF files throughout this module.

src/GridInfoType.f90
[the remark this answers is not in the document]
GreyEvenson-NOAA: I revised to put everything on one line. However, I'm not attached to this approach to styling -- I was adopting from the the ReadNamelist subroutine. What do you think? Should we put if/then statements on separate lines? There's a lot of if/then statements in this file so the file would end up looking fairly different after the change.

GreyEvenson-NOAA: I don't see a strong reason for lat and lon to be 2D. Having them as 2D may have a benefit of consistency as most model variables are either scalar or 2D/3D but I don't know if that is sufficient reason. Note that @hellkite500 in #77 alluded to maybe changing these to 1D to better conform to `bmi/bmi_grid.f90`. I wonder if we should just go ahead and make the change? Any/all input is welcome here.

With regards to the type of grids that we'd expect, I was working with the assumption that we'd only be seeing uniform rectilinear. I'd guess that I made this assumption because this is the only grid type listed in the `noah_grid_type` subroutine. However, it would be good to check my assumption with @SnowHydrology or @hellkite500. Can anyone provide a definitive answer about the types of grids that we will be working with?

test/analysis/create_dummy_netcdf.py
GreyEvenson-NOAA: Yeah, I've always hated dealing with backlashes!

src/GridInfoType.f90
nmizukami: Sorry, your can discard this. I see start of if clause (line 238), and else (in 264). maybe line238 was hiding (git comparison does not show all the lines)

nmizukami: Another comment...    is `all(read_lat(iy)/=this%lat())` supposed to use `any`.  I thought I put this comment somewhere, but maybe I did not.

src/GridInfoType.f90
andywood: I think the flexibility provided by the 2d grid seems useful (eg allows modeling on equal area grids which are used more at high latitudes) or interfacing with atmos models that often run on non-geographic (decimal degree grids).  Or even handling some ragged array situations. Also the code one-liner style seems fine esp. when clustered in a block (so it's not an isolated oddball format appearance).  it's concise and we use it in SUMMA a lot.

src/GridInfoType.f90
GreyEvenson-NOAA: Nice catch. I made the change!

src/GridInfoType.f90
GreyEvenson-NOAA: Okay, I'll leave lat and lon as 2D grids. Appreciate you sharing your thinking on this @nmizukami and @andywood. It was informative.

GreyEvenson-NOAA:
Okay, @nmizukami, I think we're ready for you to have another look. I'll review another review

src/GridInfoType.f90
hellkite500: > I think the flexibility provided by the 2d grid seems useful
@andywood Can you expand a little bit on the representation of a 2D grid that would require the coordinate arrays to be both be 2D as well?  `lat(:), lon(:)` is still a 2D grid, assuming each array is indexed by the same index.  What would the index/coordinates look like in a generic `X(:,:), Y(:,:)` representation?

src/GridInfoType.f90
nmizukami: another way would be `(.not. all(read_lon(ix) == this%lon(ix,:)))`.  maybe either way is ok I think.

src/GridInfoType.f90
nmizukami: Yes, with 2D lat/lon grid, can use both rectilinear grid and curvilinear grid. I think in the future it would be necessary to have netcdf 2D lat/lon reading routine (if not exist) for curvilinear (does not have to be done in this PR, I think)

nmizukami (approved):
Hi Grey,  I think I am fine with the changes made. thank for letting me review!

nmizukami:
I just approved (just in case i needed it?)

src/GridInfoType.f90
hellkite500: is the curvilinear grid just a different index into the `X,Y` arrays?  If so, that *could* be implemented as simply an index array?

nmizukami: curvilinear grid can have 2D lat/lon but just not regular interval (Lat/lon values in a box in curvilinear grid are not the same across the row/col). so looks like this (https://www.ncl.ucar.edu/Applications/Images/dataonmap_grid_3_lg.png).  If the data is coming from e.g., WRF, it would be likely like this.  I have WRF domain netcdf I could send you to see how lat/lon 2D variables looks like. 

I think it is good to keep 2D lat/lon variables as is but how to populate 2D lat/lon would be different, depending on which type of grid, curvilinear that come from 2D lat/lon variables or rectilinear that may come from 1D lat /lon (which is currently implemented).

PR #87, opened 2023-09-21
Noah-OM gridded namelist.input refactor

GreyEvenson-NOAA:
# PURPOSE
This PR refactors `namelist.input` and the module responsible for reading that file. This PR adds required NetCDF variable, dimension, and attribute names to `namelist.input `that were previously hardcoded.

# ADDITIONS
None

# CHANGES

## namelist.input
The following variable was added to `namelist.input`:

- `attributes_filename`, which is the directory/name of NetCDF file holding gridded model inputs

The following variable were removed from namelist.input (because they are now read-in via NetCDF file or derived from NetCDF inputs):
- `terrain_slope`
- `azimuth`
- `sfctyp`
- `soilcolor`

Note that the `vegtyp` and `isltyp` variables were previously removed in PRs #74 and #80, respectively.

## NamelistRead.f90
The `NamelistRead` module was refactored to facilitate reading the refactored `namelist.input` file.

## DomainGridType.f90
`domaingrid_type%IST` (i.e., the soil vs. lake indicator variable) was removed from the `domaingrid_type%InitTransfer` procedure. 

This PR uses the NetCDF-provided input variable `vegtyp` (land use) and the variable `parameter_type%ISWATER` (from `/parameters/MPTABLE.TBL`) to derive `domaingrid_type%IST` rather than require `sfctyp` as an additional gridded input variable. However, we cannot set `domaingrid_type%IST` here because `parameters_type%ISWATER` has not yet been set -- and won't be until `parametersgrid_type%paramRead` is called. Thus, `IST` was removed from `domaingrid_type%InitTransfer`.

## RunModule.f90

This PR initializes `domaingrid_type%IST` in `initialize_from_file` as:
```
      domaingrid%IST(:,:) = 1
      where (domaingrid%vegtyp == parametersgrid%ISWATER) domaingrid%IST = 2
```

# TESTING

I re-compiled the model, re-executed the model, and re-executed /test/analysis/compare_outputnc.py and got expected resutls: grid cell (1,1) passed all tests while all other cells did not.

I also did a crude test to verify that the PR correctly initializes `domaingrid_type%IST` by adding a print statement immediately after `IST`'s initialization, i.e.:

Screenshot 2023-09-22 at 9 20 08 AM

I then modifyed `/data/netcdf_input.nc` such that all grid cells had a `vegtyp` value of 1 except for cell (1,1). Cell (1,1) was assigned a `vegtyp` value of 17, which equals `ISWATER` if `veg_class_name = "MODIFIED_IGBP_MODIS_NOAH"`. I then re-compiled the model with the included print statement and re-executed with the modified `/data/netcdf_input.nc` file and got the following (correct) output:

Screenshot 2023-09-21 at 3 38 59 PM


GreyEvenson-NOAA:
@SnowHydrology: @nmizukami was kind enough to volunteer to review this. However, I wanted to call your attention to this PR because (1) it refactors `namelist.input`, which seems important, and (2) it does not require that the `IST` variable (i.e., soil vs lake indicator variable) be read-in via NetCDF -- the PR instead initializes `IST` via `vegtyp` (land use). Let me know if you have thoughts on this.

SnowHydrology (commented):
@GreyEvenson-NOAA, this looks good to me and the model runs correctly.

@nmizukami and @andywood, how do you feel about the variable names for the netcdf files in the namelist? Is that a good place or should they go somewhere else?

andywood (commented):
my main two reactions are:
* I would avoid a generic name like 'netcdfin_filename' that doesn't given any sense of what the file contains, versus something descriptive/intuitive, eg 'param_filename' or 'static_input_filename' or 'attrib' etc.  I like the way summa separates parameters from attributes (all static but different in nature), where the param file data will get altered more often but the geophys attribs don't change unless the spatial/vertical configuration of the model changes.  

* as a rule, I don't like bulky config files, especially if they contain things that a user would not need to see or change very often (or at all).  So I would go back to just specifying the netcdf filename here, while hardwiring the convention for what variables in the netcdf file need to be named.  It cuts down on code and config file bloat.  many models do this (fixed input var names) because it's not an onerous user requirement that the user prepare the input file following some conventions:  eg, vegtyp, isltyp, soilcolor, slope, aazimuth, latitude, longitude, dx, dy (and no caps on Latitude, Long to be consistent).  we do want to allow users some control and flexibility but it's best to focus it on things that make a big difference either for them or their modeling options -- my opinions is that they will be fine if they don't have their own special input name for latitude.  Having some reasonable conventions can also make analysis across input files and configurations easier, as more things can be assumed.  

btw, nice to see some summa-style code and missingType variables included, e.g., 
if(name_att_dy /= stringMissing) then; this%name_att_dy = name_att_dy; else; write(*,'(A)') 'ERROR: required entry name_att_dy not found in namelist'; stop; end if
this code looks straight out of summa, possibly through naoki?

run/namelist.input
```diff
@@ -51,8 +45,19 @@
  nsnow            = 3               ! number of snow levels
  nveg             = 20              ! number of vegetation types
  croptype         = 0               ! crop type (0 = no crops; this option is currently inactive)
- sfctyp           = 1               ! land surface type, 1:soil, 2:lake
- soilcolor        = 4               ! soil color code
+/
+
+&netcdf_input
```
nmizukami, line 50: This could be renamed.  e.g., attributes or geophysical_attrs (maybe too long?)

run/namelist.input
```diff
@@ -51,8 +45,19 @@
...
+ netcdfin_filename  = "../data/netcdf_input.nc"  ! directory/name of NetCDF file holding gridded model inputs
```
nmizukami, line 51: could rename `netcdfin_filename` to `attribute_netcdf` (or something like this)

src/NamelistRead.f90
```diff
@@ -139,25 +144,40 @@ subroutine ReadNamelist(this, namelist_file)
...
                                  dynamic_vic_option,dynamic_veg_option,snow_albedo_option,&
                                  radiative_transfer_option,sfc_drag_coeff_option,canopy_stom_resist_option,&
                                  crop_model_option,snowsoil_temp_time_option,soil_temp_boundary_option,&
                                  supercooled_water_option,stomatal_resistance_option,&
                                  evap_srfc_resistance_option,subsurface_option
-    namelist / structure       / nsoil,nsnow,nveg,croptype,sfctyp,soilcolor
+    namelist / structure       / nsoil,nsnow,nveg,croptype
     namelist / initial_values  / zsoil,dzsnso,sice,sh2o,zwt    
-    
+    namelist / netcdf_input    / netcdfin_filename,name_var_vegtyp,name_var_isltyp,name_var_soilcolor,&
```
nmizukami, line 178: consider modifying namelist group name and variable name for netcdf name

nmizukami (commented):
Hi Grey,  

After looking at the namelist example (`namelist.input`), I agree with Andy's first comment on variable name for netcdf - `netcdfin_filename`.  This can be more specific name (I think this is easy to fix in this PR).

For the second comment from Andy,  if you make the NetCDF variable names hardcoded, you would need to add the lines to check if the variable exist in the netCDF given the name, using `nf90_inq_varid` (maybe another netCDF function?), and tell the user that netCDF variables need to be xxxx.  

or you provide the hardcoded default name and allow the user to overwrite them if user provide something other than default name in netCDF for some reason.  But, I am not sure if you can do optional namelist variables. 

For now, netCDF input variables are not too many, so looks ok, but if the list of variables increase, it will become unreasonable for the user to input them in the namelist. So I am leaning having default names.

src/RunModule.f90
```diff
@@ -238,6 +238,8 @@ SUBROUTINE initialize_from_file(model, config_filename)
       do ii = 1, namelist%nsoil
         domaingrid%zsnso(:,:,ii) = namelist%zsoil(ii)
       end do
```
nmizukami, line 240: might be good to comment what is 1 and 2?   I think initializing lake or waterbody from land cover information makes sense to me (i guess this is called vegtype, so feel strange? but i think it is ok.  I guess noah traditionally uses vegetation type rather than land cover type?)

run/namelist.input
andywood: I think the netcdf aspect is not too significant ... versus that it's a 'filename' and what it contains.  A hybrid could be used, too, eg 'attrib_nc_file' or something -- & you're right, these are attributes, not params.

run/namelist.input
andywood: @naoki, I'd think a inq_varid call is needed regardless of whether the varname is user specified or not. 
you can get away without it but then can't trap the explicit error and message it as clearly.

GreyEvenson-NOAA:
I appreciate you all taking the time to look at this, @SnowHydrology, @andywood, and @nmizukami. 

I'll revise this PR to rename the netcdfin_filename to attrib_nc_file or something similar. I'll also hardcode the NetCDF variable, attribute, and dimension names and make sure that we have sufficient checks and error messages in case the user provides unexpected names/values. I'll additionally make sure to address anything else that you've identified in the comments above.

@andywood: regarding that SUMMA-style code, whomever wrote the NamelistRead subroutine gave us that gift. That's where I lifted it from!

Thanks again, all! Appreciate it.

SnowHydrology (changes requested):
Thanks @nmizukami and @andywood for your feedback on the namelist. @GreyEvenson-NOAA, let me know when you've integrated their changes and I'll review again. Thanks!

run/namelist.input
GreyEvenson-NOAA: Given this discussion, would you folks suggest that we rename the `NetCDFVarsType` module and the derived types that are defined within?  For example, the `NetCDFVarsType` module could be renamed the `BasinAttributesType` module or the `DomainAttributesType`? `netcdfvars_type` could be renamed `basinattributes_type` or `domainattrib_type`? Let me know if you have thoughts @andywood @nmizukami @SnowHydrology. I will default to making no change the module and derived type names.

andywood: definitely -- the type should be more related to its significance (eg params, attributes, forcings etc) than the filetype they come from.  good call :).  In this case I thinking we're dealing with attributes, so an attrib_type would be good.  if there's a scale-related scope (eg global, basin, local) that could be part of the type name too.

andywood:
> I appreciate you all taking the time to look at this, @SnowHydrology, @andywood, and @nmizukami.
> 
> I'll revise this PR to rename the netcdfin_filename to attrib_nc_file or something similar. I'll also hardcode the NetCDF variable, attribute, and dimension names and make sure that we have sufficient checks and error messages in case the user provides unexpected names/values. I'll additionally make sure to address anything else that you've identified in the comments above.
> 
> @andywood: regarding that SUMMA-style code, whomever wrote the NamelistRead subroutine gave us that gift. That's where I lifted it from!
> 
> Thanks again, all! Appreciate it.

Could have been me pulling that in -- a sad part of fractured attention to too many projects is that I don't remember!

src/RunModule.f90
GreyEvenson-NOAA: Probably a good idea -- I added comments to describe IST value meaning.

Yes, the vegtyp = land use did seem odd to me initially. I don't know the backstory but maybe because the model was initially aiming to get vegetation phenology correct rather than human activities such as agriculture (?).

src/NamelistRead.f90
GreyEvenson-NOAA: This got changed to: `namelist / attributes        / attributes_filename`. Let me know if you think the names could be improved.

run/namelist.input
GreyEvenson-NOAA: I changed lots of names in revising the PR:

`NetCDFVarsType.f90` -> `AttributesType.f90`
`netcdfvars_type` -> `attributes_type`
`netcdfin_filename` -> `attributes_filename`

However, I'm not married to these. I'd be happy to change them again if folks think it would be helpful.

andywood:
This all looks good to me, Grey :)

GreyEvenson-NOAA:
@SnowHydrology, I believe we're ready for you to have another look. @andywood seemed to approve of the changes. @nmizukami, let me know if you see anything else we can improve upon.

Thanks again for all your time in reviewing this!

data/netcdf_input.nc
GreyEvenson-NOAA, line 1: The latitude and longitude dimensions and variables are now lower-case to be consistent with names.

test/analysis/create_dummy_netcdf_singlefile.py
GreyEvenson-NOAA, line 1: Modified the script to create the input netcdf file with lower case latitude and longitude names.

src/AttributesType.f90
```diff
@@ -0,0 +1,492 @@
...
+    !----------------------------------------------------------------------------
+    select type (attributes_2d)
+    type is (attributes_2dint_type)
+      status = nf90_get_var(ncid = ncid,varid = varid, values = attributes_2d%data)
+      if (status /= nf90_noerr) then; write(*,*) 'Unable to read variable ''',trim(varname),''' from ''',trim(filename),''''; stop ":  ERROR EXIT"; end if
+      if(attributes_2d%data(1,1) == integerMissing) then; write(*,*) 'ERROR : problem reading ''',trim(varname),''' from ''',trim(filename),''''; stop; end if
+    type is (attributes_2dreal_type)
+      status = nf90_get_var(ncid = ncid,varid = varid, values = attributes_2d%data)
+      if (status /= nf90_noerr) then; write(*,*) 'Unable to read variable ''',trim(varname),''' from ''',trim(filename),''''; stop ":  ERROR EXIT"; end if
+      if(attributes_2d%data(1,1) == realMissing) then; write(*,*) 'ERROR : problem reading ''',trim(varname),''' from ''',trim(filename),''''; stop; end if
```
nmizukami, line 427: One last minor thing here...    I think maybe add default (unlikely to come here in this situation, but just for completeness and good practice)
 
```
  class default
     write(*,*) 'error message'; stop ": ERROR EXIT"
  end select
```

also the same for previous select type clause.

nmizukami (approved):
Hi Grey,  I put one minor comment but I don't see anything else to add.  thank you for incorporating this!

src/AttributesType.f90
GreyEvenson-NOAA: Nice catch. I just added the `class default` to both of the `select type` statements here.

SnowHydrology (approved):
Model runs with expected values and unit tests pass

Issue #97, opened 2024-02-07
Check for consistency in land cover and soil


SnowHydrology:
# Current behavior

You can run Noah-OM with nonsensical, inconsistent soil and land cover values. As we saw in https://github.com/NOAA-OWP/DMOD/issues/472#issuecomment-1893759701, this can lead to model errors.

# Expected behavior

Noah-OM should not initialize if inconsistent values of land cover and soil are provided.

# Suggested changes

I've flagged an issue on the [hydrofabric repo](https://github.com/NOAA-OWP/hydrofabric/issues/35), but we should also implement model-level sanity checks. These will stop initialization and throw an error message if land cover and soil are inconsistent.

Specifically, Noah-OM should check the values of `isltyp` and `vegtyp` for consistency. I provide example category lumpings in the linked hydrofabric issue above.

PR #100, opened 2024-02-22
Add check for consistency in land cover and soil

GreyEvenson-NOAA:
# Purpose

This PR fixes issue #97. 

# Additions

Added land cover and soil consistency checks to `initialize_from_file` in `RunModule.f90`. These checks throw an error and stop the program if any cell is initialized with the following conditions:

- land use type != water and soil type = water
- land use type = water and soil type != water
- land use type != ice and soil type = ice
- land use type = ice and soil type != ice

# Testing

I manually/hardcoded problematic values for vegtyp and isltyp to produce errors:

`where (domaingrid%mask.eq.1) domaingrid%vegtyp = parametersgrid%ISWATER`
`where (domaingrid%mask.eq.1) domaingrid%isltyp = 1`
returns => `ERROR: one or more grid cells have a water land cover (vegtyp = 16) and a non-water soil type (isltyp != 14)`

`where (domaingrid%mask.eq.1) domaingrid%vegtyp = 1`
`where (domaingrid%mask.eq.1) domaingrid%isltyp = 14`
returns => `ERROR: one or more grid cells have a non-water land cover (vegtyp != 16) and a water soil type (isltyp = 14)`

`where (domaingrid%mask.eq.1) domaingrid%vegtyp = parametersgrid%ISICE`
`where (domaingrid%mask.eq.1) domaingrid%isltyp = 1`
returns => `ERROR: one or more grid cells have an ice land cover (vegtyp = 24) and a non-ice soil type (isltyp != 16)`

`where (domaingrid%mask.eq.1) domaingrid%vegtyp = 1`
`where (domaingrid%mask.eq.1) domaingrid%isltyp = 16`
returns => `ERROR: one or more grid cells have a non-ice land cover (vegtyp != 24) and an ice soil type (isltyp = 16)`

# Notes

- Implemented checks using `where` statements to avoid iterating over the grid
- Checks were implemented with hardcoded water and ice values for soil type. To avoid using hardcoded soil type values, we could require that the user provide `ISWATER` and `ISICE` values for the utilized soil classification scheme (as the user already must provide for the utilized land cover classification scheme).


src/RunModule.f90
```diff
@@ -245,15 +246,28 @@ SUBROUTINE initialize_from_file(model, config_filename)
...
+      if(any(err_grid.eq.1)) then; write(*,'(A,i2,A)') 'ERROR: one or more grid cells have a non-water land cover (vegtyp != ',parametersgrid%ISWATER,') and a water soil type (isltyp = 14)'; stop; end if
+      where (domaingrid%mask.eq.1.and.domaingrid%vegtyp.eq.parametersgrid%ISWATER.and.domaingrid%isltyp.ne.14) err_grid = 1
+      if(any(err_grid.eq.1)) then; write(*,'(A,i2,A)') 'ERROR: one or more grid cells have a water land cover (vegtyp = ',parametersgrid%ISWATER,') and a non-water soil type (isltyp != 14)'; stop; end if
+      where (domaingrid%mask.eq.1.and.domaingrid%vegtyp.ne.parametersgrid%ISICE.and.domaingrid%isltyp.eq.16) err_grid = 1
+      if(any(err_grid.eq.1)) then; write(*,'(A,i2,A)') 'ERROR: one or more grid cells have a non-ice land cover (vegtyp != ',parametersgrid%ISICE,') and an ice soil type (isltyp = 16)'; stop; end if
+      where (domaingrid%mask.eq.1.and.domaingrid%vegtyp.eq.parametersgrid%ISICE.and.domaingrid%isltyp.ne.16) err_grid = 1
+      if(any(err_grid.eq.1)) then; write(*,'(A,i2,A)') 'ERROR: one or more grid cells have an ice land cover (vegtyp = ',parametersgrid%ISICE,') and a non-ice soil type (isltyp != 16)'; stop; end if
```
PhilMiller, line 269: The code could de-allocate `err_grid` here, since it's not going to be used later in the simulation

drakest123 (changes requested):
If possible, report problematic grid cell locations rather than the fact that there is/are problematic grid cells.

src/RunModule.f90
GreyEvenson-NOAA: Appreciate the catch, @PhilMiller

GreyEvenson-NOAA:
Thanks for the review @drakest123. I went ahead and add locational information to the error messages.

GreyEvenson-NOAA:
@SnowHydrology Do you think this PR's checks are too restrictive? Here, we're requiring grid cells that have `ice` land cover to have `ice` soils as well.

SnowHydrology (changes requested):
This looks good with one exception: using `stop` prevents the framework from returning useful error information. Instead, I'd recommend using `return` and then having the initialize routine return `BMI_FAILURE`. See examples [here](https://github.com/SnowHydrology/noah-owp-modular/blob/f939e07ba45397bbc697f83cedbcbb19ad4c2256/bmi/bmi_noahowp.f90#L272) and [here](https://github.com/SnowHydrology/noah-owp-modular/blob/f939e07ba45397bbc697f83cedbcbb19ad4c2256/src/EnergyModule.f90#L309).

GreyEvenson-NOAA:
Note: I'm leaving the `stop` statements as-is with the assumption that the issue will be corrected by PR #108.

src/RunModule.f90
```diff
@@ -95,8 +95,10 @@ SUBROUTINE initialize_from_file(model, config_filename)
     
     type (noahowpgrid_type), intent (out)   :: model
     character(len=*), intent (in)           :: config_filename    ! config file from command line argument
-    integer                                 :: forcing_timestep         ! integer time step (set to dt) for some subroutine calls
-    integer                                 :: ii, ix, iy
+    integer                                 :: forcing_timestep   ! integer time step (set to dt) for some subroutine calls
+    integer                                 :: ii                 ! indices
+    integer, allocatable, dimension(:,:)    :: err_grid           ! to catch land use vs. soil inconsistencies
+    integer, dimension(2)                   :: err_indices        ! to hold ix, iy for err_grid
```
GreyEvenson-NOAA, line 101: Note: it may be preferable to move these error-related objects to the ErrorCheckModule if PR #108 elects to make greater use of that module for error checking and messaging purposes.

PR #105, opened 2024-03-19
don't initialize ngen submodule after copying in local code

### LINKED ISSUE #104: GH workflow builds against ngen submodule commit, not current commit
https://github.com/NOAA-OWP/noah-owp-modular/blob/05aec159a872fe4ba5b6a490c4ed91878b2396cb/.github/workflows/ngen_integration.yaml#L95

Just noticed that the workflow here is using the ngen build-submodule action to build the NOM library to test the integration with.  This will cause the runner to use the pinned ngen submodule commit to build form, not the current code.  This likely isn't the intended behavior here.

Can either toy with the ngen submod build action to try to get the correct code/pr-branch updated in the runner's ngen clone, or simply build the ngen lib from source (would be easier once #103 is complete).  Could also hack somewhere in the middle and use the checked out code to overwrite the dir and try to re-build via a second action step.

Either way, something to think about for getting a local integration test setup.

hellkite500:
This PR, dependent on noaa-owp/ngen#769, would allow the integration test to use the current checked out code in place of the submodule checkout by using the new "initialize" input.  This also passes custom cmake build flags along to the submodule build action to configure the build.

This allows the ngen build to use the static submodule commits (which have hopefully been tested) and only use the current NOM PR code to run the test with.  Customization of other components is possible, but not in scope here and may not be a good idea in general.

Closes #104

### DIFF
--- a/.github/workflows/ngen_integration.yaml
+++ b/.github/workflows/ngen_integration.yaml
@@ -98,6 +98,8 @@ jobs:
         with:
           mod-dir: " extern/noah-owp-modular/"
           targets: "surfacebmi"
+          initialize: "false"
+          cmake-flags: "-DNGEN_IS_MAIN_PROJECT:BOOL=ON"
 
       - name: Run Ngen Test
         run: |


hellkite500:
Integration tests will likely fail until the action is updated in ngen

madMatchstick (approved):
Integration tests now pass, thanks @hellkite500

PR #106, opened 2024-04-19
Avoid relying on non-standard Fortran syntax extension for unary negation

PhilMiller:
The code as written leads gfortran to produce this warning:

```
Warning: Extension: Unary operator following arithmetic operator (use parentheses)
```


PhilMiller:
Many of the cases touched in my change could maybe be more clearly rewritten by replacing addition with subtraction, or reciprocal by exponentiation with division. I opted to take a purely syntactic approach. I'll defer to the Noah developers on how they'd like arithmetic to be written out in the code.

GreyEvenson-NOAA (approved):
Looks good. Tested with `-Wall`.

PR #115, opened 2024-09-03
reverting evapotrans units back to m/s

SnowHydrology:
This PR reverts the earlier modification to the units of `evapotrans` that make it incompatible with other NextGen modules.


hellkite500 (approved):

hellkite500:
My only suggestion is to consider whether `ETRAN` and `ECAN` need the same treatment.

andywood:
@Nels good suggestion -- both are fluxes and hence good candidates for units of L/T or M/T (ie kg-m s-1).  ETRAN is a current output that could be switched, and even if ECAN is not (yet), that may be of interest later.

SnowHydrology:
I agree, @hellkite500 and @andywood. I'll open an issue for future updates.

PR #132, opened 2026-07-10
Clear string/integer roundtripping from date/time processing

### LINKED ISSUE #131: Excessive string parsing and trimming in the main loop
https://github.com/NOAA-OWP/noah-owp-modular/blob/0ff055efd93c37f5bce83255f06180e2e779cea6/src/UtilitiesModule.f90#L31-L37

There's a significant performance issue here in the way that `geth_newdate` and `calc_declin` process dates for the __per timestep__ declination calculation.  These get through the `UtilitiesMain` subroutine [here](https://github.com/NOAA-OWP/noah-owp-modular/blob/0ff055efd93c37f5bce83255f06180e2e779cea6/src/RunModule.f90#L292) in the `solve_noahowp` subroutine.

The performance penalty is parsing a string date, computing a new string date, passing those string components to `calc_declin`, just to have it convert them back to integer components for its calculation.

Ironically enough, the intermediate integer date components are already held in `geth_newdate`, and so one possible option is to simply output those directly into `calc_declin`, but that only reduces about half of the string manipulation.

Note that `geth_newdate` is also used by the `read_forcing_text` function, so it probably needs to remain as is unless we want to touch that as well, which I'm not inclined to do at this point as that is not on any critical path (I think that's really only a helper for standalone forcing...)

Parsing and holding the integer `startdate` components once, and then computing the new components from those would greatly reduce the string handling on the main loop.

A revised/supplemental `calc_declin` could take those components directly, and avoid this call
https://github.com/NOAA-OWP/noah-owp-modular/blob/0ff055efd93c37f5bce83255f06180e2e779cea6/src/UtilitiesModule.f90#L869
by directly computing the day of the year.

Care should be taken to account for the various rollovers in dates as `geth_newdate` does, as well as the day of year edges shown in `geth_idts`.

PhilMiller:
Fixes #131


PhilMiller:
I recommend reviewing by skimming the sequence of commit messages to understand the progression, and then reading the commits in sequence.

We could choose to do without the last two commits. They represent an ancillary cleanup/follow-through on the spirit of the primary task set by 131.

aaraney:
Thanks for staging these commits to simplify the review process! I had a quick skim and nothing sticks out to me.

test/datetime_test.f90
```diff
@@ -0,0 +1,420 @@
...
+     , declin_case("julian should be 1-based per docs" ,  .false.,  2024,   1,   1,   0,   0,   0,  40.0,  -105.0,    0.0,      0.0,  -0.068059,    -0.068059,       366,    1.000000,  1.0e-4) &
+     , declin_case("julian should include minutes"     ,  .false.,  2024,   3,   5,   6,  30,   0,  40.0,  -105.0,    0.0,      0.0,  -0.823714,    -0.823714,       366,   64.270833,  1.0e-4) &
+     ]
+
+  !                  label                          ,   epass,    yr,  mo,  dy,  hr,  mi,  dminutes,   lat,     lon,  slope,  azimuth,     ecosz,  ecosz_horiz,  eyearlen,     ejulian,     tol
+  type(composed_case), parameter :: composed_cases(*) = [ &
+       composed_case("advance 1998-01-01 +27000min" ,  .true.,  1998,   1,   1,   0,   0,     27000,  40.0,  -105.0,    0.0,      0.0,  0.471150,     0.471150,       365,   18.750000,  1.0e-4) &
+     , composed_case("advance 1998-01-01 +782940min",  .true.,  1998,   1,   1,   0,   0,    782940,  40.0,  -105.0,    0.0,      0.0,  0.863534,     0.863534,       365,  178.708333,  1.0e-4) &
+     ]
```
hellkite500, line 182:
```suggestion
     , composed_case("advance 1998-01-01 +782940min",  .true.,  1998,   1,   1,   0,   0,    782940,  40.0,  -105.0,    0.0,      0.0,  0.863534,     0.863534,       365,  178.708333,  1.0e-4) &
     , composed_case("div-3600: 3600 is leap, Feb28 +2 days -> Mar1",  .true.,  3600,   2,  28,   0,   0,      2880,  40.0,  -105.0,    0.0,      0.0,  0.109952,     0.109952,       366,   60.000000,  1.0e-4) &
     ]
```

This suggested test shows the discrepancy of the `%3600` inconsistency.

src/UtilitiesModule.f90
```diff
@@ -805,26 +89,131 @@ integer function nmdays(hdate)
...
+  ! ---------------------------------------------------------------------
+  ! Integer-component date/time routines (issue #131). These carry dates
+  ! as integer components (year, month, day, hour, minute) so the model
+  ! main loop can avoid per-timestep string parsing and formatting.
+  ! ---------------------------------------------------------------------
+
+  ! Day number of a Gregorian civil date relative to 1970-01-01
+  ! (negative for earlier dates). Days-from-civil algorithm of
+  ! Howard Hinnant, "chrono-Compatible Low-Level Date Algorithms".
+  integer function days_from_civil(yr, mo, dy)
```
hellkite500, line 89: This algorithm doesn't consider the `%3600` leap year case the same as `nfeb` and the inline year length calculation in the declination computation.  The suggested test case below demonstrates that as is there is disagreement.

src/UtilitiesModule.f90
```diff
@@ -866,10 +248,6 @@ SUBROUTINE calc_declin (nowdate, &                             ! in
     endif
```
hellkite500, line 228:
```suggestion
yearlen = 365 + (nfeb(iyear) - 28)
```
Similar to the previous comment, this isn't consistent with the days_from_civil. Either remove the %3600, or use the suggestion with a fixed `nfeb` so there is a consistent representation of leap years.

src/UtilitiesModule.f90
hellkite500, line 62:
```suggestion
```
This will cause an inconsistency with the `days_from_civil` day of the year calculation which use a standard Gregorian `%4/%100/%400` for computing leap years. To keep the two consistent and not be off by one for every day every 3600 years, we should drop this here.

hellkite500 (changes requested):
My local testing revealed one inconsistency that I think should be addressed (although we my not hit this case in our typical uses, it is simple enough to ensure consistency)

src/UtilitiesModule.f90
```diff
@@ -805,26 +89,131 @@ integer function nmdays(hdate)
...
+       doy = (153*(mo + 9) + 2)/5 + dy - 1
+    end if
+    doe = yoe*365 + yoe/4 - yoe/100 + doy              ! [0, 146096]
+    days_from_civil = era*146097 + doe - 719468
+  end function days_from_civil
+
+  ! Signed difference newdate - olddate in minutes, for 12-char
+  ! YYYYMMDDHHmm date strings. Each string is parsed once (integer
+  ! components), unlike the retired geth_idts which validated and
+  ! re-parsed both dates on every call.
```
hellkite500:
```suggestion
  ! Signed difference newdate - olddate in minutes, for 12-char
  ! YYYYMMDDHHmm date strings.
```
Drop the dead/deleted code reference

src/UtilitiesModule.f90
```diff
@@ -805,26 +89,131 @@ integer function nmdays(hdate)
...
+       mo = mp + 3
+    else
+       mo = mp - 9
+    end if
+    yr = yoe + era*400
+    if (mo <= 2) yr = yr + 1
+  end subroutine civil_from_days
+
+  ! Advance a date/time by a signed number of minutes. O(1) day-count
+  ! arithmetic, unlike geth_newdate's day-at-a-time loop.
```
hellkite500:
```suggestion
  ! Advance a date/time by a signed number of minutes. 
```
Another dead code reference

src/UtilitiesModule.f90
```diff
@@ -805,26 +89,131 @@ integer function nmdays(hdate)
     endif
   end function nmdays
```
hellkite500:
```suggestion
```

I think `nmdays` is now dead code that was only used in the previous date time handling, it can drop with this PR as well.

test/datetime_test.f90
```diff
@@ -0,0 +1,420 @@
...
+     real    :: ejulian                   ! expected floating-point day of year
+     real    :: tol                       ! tolerance for real comparisons
+  end type composed_case
+
+  ! ---- implementations under test -------------------------------------------
+  ! Every table runs against each implementation listed for it. A string
+  ! adapter is removed only when the string routine it wraps is removed
+  ! from the codebase; while both implementations exist, both are covered.
+
+  integer, parameter :: IMPL_STRING = 1, IMPL_COMPONENT = 2
```
hellkite500: The `IMPL_STRING` isn't used after the migration completed...

src/UtilitiesModule.f90
hellkite500: Should this now be 0 based, and the `julian should be 1-based per docs` test case re-considered?

robertbartel: > ... not be off by one for every day every 3600 years ...

As a sanity check, do we really care about this @hellkite500?  I'd imagine there'd be a larger chance of accidentally introducing a human-error-based bug making a change than actually running a simulation over year 0, 3600, etc.  Am I missing something?

src/UtilitiesModule.f90
```diff
@@ -805,26 +89,131 @@ integer function nmdays(hdate)
...
+
+    integer, parameter :: cum(12) = (/ 0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334 /)
+
+    day_of_year = cum(mo) + dy - 1
+    if (mo > 2) day_of_year = day_of_year + (nfeb(yr) - 28)
+  end function day_of_year
+
+  ! Solar geometry from integer date/time components: the body of
+  ! calc_declin minus all string parsing. iday is the day-of-year offset
+  ! as computed by day_of_year().
```
hellkite500:
```suggestion
  ! Solar geometry from integer date/time components.
  ! iday is the day-of-year offset as computed by day_of_year().
```

test/datetime_test.f90
```diff
@@ -0,0 +1,420 @@
...
+! thin adapter and reports PASS/FAIL.
+!
+! Each table runs against every implementation listed for it (see the impl
+! lists below): the component routines (advance_datetime, day_of_year,
+! calc_declin_components) and, where a string routine still exists in the
+! codebase (geth_idts), a string adapter alongside.
+!
+! Expected values were derived independently of the implementation:
+! calendar arithmetic with Python datetime (proleptic Gregorian), solar
+! geometry from the documented formulas evaluated in double precision.
```
hellkite500: If this is intended to stay as *active* test coverage for these components, then this should reflect the *current* state and use of these tests in its final form, and should at least not refer to the dead code dropped by the migration.

test/datetime_test.f90
```diff
@@ -0,0 +1,420 @@
...
+  ! ---- implementations under test -------------------------------------------
+  ! Every table runs against each implementation listed for it. A string
+  ! adapter is removed only when the string routine it wraps is removed
+  ! from the codebase; while both implementations exist, both are covered.
+
+  integer, parameter :: IMPL_STRING = 1, IMPL_COMPONENT = 2
+  character(len=11), parameter :: impl_name(2) = (/ "[string]   ", "[component]" /)
+
+  ! All string routines are retired (geth_newdate and calc_declin in phase
+  ! 3c, geth_idts in phase 5), so only the component implementation remains.
```
hellkite500: Some planning/iteration comments left  over here.

src/UtilitiesModule.f90
```diff
@@ -23,741 +23,25 @@ SUBROUTINE UtilitiesMain (itime, domain, forcing, energy)
...
     integer  :: idt ! change in time since beginning of run (in minutes)
+    integer  :: now_year, now_month, now_day, now_hour, now_minute ! current date/time components
     idt = itime * (domain%dt / 60)
 
-    ! calculate current 'nowdate' from start date + integer length of run to current time
-    call geth_newdate(domain%startdate, idt, &  ! in
-                      domain%nowdate)           ! out
+    ! calculate current date components from start date components (parsed
+    ! once at init) + integer length of run to current time; no string
+    ! parsing on this path (issue #131)
```
hellkite500:
```suggestion
    ! calculate current date components from start date components (parsed
    ! once at init) + integer length of run to current time;
```

src/UtilitiesModule.f90
```diff
@@ -805,26 +89,131 @@ integer function nmdays(hdate)
...
 
-  SUBROUTINE calc_declin (nowdate, &                             ! in
-                          latitude, longitude, slope, azimuth, & ! in
-                          cosz, cosz_horiz, yearlen, julian)                 ! out
-  !---------------------------------------------------------------------
+  ! ---------------------------------------------------------------------
+  ! Integer-component date/time routines (issue #131). These carry dates
+  ! as integer components (year, month, day, hour, minute) so the model
+  ! main loop can avoid per-timestep string parsing and formatting.
+  ! ---------------------------------------------------------------------
```
hellkite500:
```suggestion
  ! ---------------------------------------------------------------------
  ! Integer-component date/time routines. These carry dates
  ! as integer components (year, month, day, hour, minute)
  ! ---------------------------------------------------------------------
```

robertbartel (commented):
I'm willing to approve but want to leave time for further discussion on the year-3600 item if others wish.  I'll note a couple of small things that caught my eye:

* The domain now tracks "start date" state values in two (non-string) places, which is less than ideal.  
* I personally think the function name `advance_datetime` carries implications of modifying state and would be better off named something slightly different.  
 
But those are nit-picky items (the second very much so) that can be left alone or deferred.

hellkite500 (commented):
Some additional cleanup/housekeeping review notes on inline documentation and maintenance (test code...)

src/UtilitiesModule.f90
hellkite500: The upstream NoahMP has been used in typical climatological scenarios, so I don't want to assume that this fork won't see the same treatment at some point, even in the context of hydrological modeling by a research community interested in these long term projections.

I would rather address this now than debug it later.

test/datetime_test.f90
PhilMiller: I had originally told Claude to disregard the 3600 cases, as a non-standard extension of the Gregorian calendar, that was well beyond our use cases' time horizon. If we care about preserving that for climatological use cases, I'll bring it back.

hellkite500: It's simple enough to just drop the non standard checks and have a documented standard Gregorian calendar representation that doesn't have a hidden inconsistency.

src/UtilitiesModule.f90
PhilMiller: The %3600 behavior is now going to be preserved throughout.

I agree your suggested implementation is nicer, but this block code is otherwise unmodified, and I'd like to keep it that way.

test/datetime_test.f90
PhilMiller: I'm going to leave the extended Gregorian calendar with the year-3600-is-not-leap behavior in place, and add an adjustment of this test to `eyearlen == 365` accordingly. Other test cases are now consistent with that.

src/UtilitiesModule.f90
PhilMiller: Perhaps we should also go through and make the documentation of Julian day consistent with the implementation. I'm disinclined toward making general code-janitorial changes that stray too far from fixing the inciting issue for this PR. We should get that fixed as the priority.

PhilMiller: OK, I asked Claude to take a look, and it came to the same conclusion you did:

> ❯ Analyze the code regarding the discrepancy in julian day being 0 vs 1 based. Which things say or do one vs the other? Are the various uses in code consistent with the physical interpretation for which it's being used?                                                                                                                                                                  

  Who says what                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                             
  Says 1-based (one place): the UtilitiesModule header comment (src/UtilitiesModule.f90:6): "JULIAN = day of year from 1 to 365 (or 366 for leap year)". That's the only place in the codebase claiming 1-based, and it's a documentation line dating from the original module.                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                             
  Does 0-based (everything else):                                                                                                                                                                                                                                                                                                                                                            
  - Producer — calc_declin_components computes julian = real(iday) + real(ihour)/24. where iday is days since Jan 1 (Jan 1 = 0). So Jan 1 00:00 → 0.0, Dec 31 non-leap → 364.x. (The retired string code did the same via geth_idts(date, "YYYY-01-01").)                                                                                                                                    
  - Consumer 1: solar longitude, same routine — SXLONG = (360/365)·(JULIAN−80) (with the +285 branch making it continuous across New Year). This pins declination = 0 at julian = 80.0.                                                                                                                                                                                                      
  - Consumer 2: phenology (InterceptionModule.f90:55-78) — LAI/SAI monthly interpolation via T = 12·DAY/yearlen, where DAY = julian (Northern Hemisphere) or MOD(julian + yearlen/2, yearlen) (Southern). Its local declaration comment says explicitly: DAY !current day of year ( 0 <= DAY < forcing%yearlen ) — this consumer documents the 0-based convention.                           
  - Init — RunModule.f90:179 sets forcing%JULIAN = 0.0 ("arbitrary"), a valid 0-based Jan 1.                                                                                                                                                                                                                                                                                                 
  - forcing%julian isn't exposed through BMI, so there's no external contract either way.                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                             
  Is each use physically consistent?                                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                             
  Phenology: yes, and it specifically wants 0-based. The interpolation maps "fraction of year elapsed" onto mid-month-anchored climatology: with 0-based DAY, Jan 1 gives T = 0 → weights 0.5/0.5 between December and January, which is exactly right for a date halfway between the mid-Dec and mid-Jan anchor points. The Southern-Hemisphere MOD(julian + yearlen/2, yearlen) also       
  assumes input in [0, yearlen). A switch to 1-based would shift every LAI/SAI value by one day — mechanically it wouldn't crash (the wraparound guards handle T = 12), but it would violate this routine's documented range.                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                             
  Declination: consistent as a matched pair, mediocre astronomically under either base. The constant 80 and the 0-based julian were inherited together from the WRF/HRLDAS lineage (WRF's JULIAN is likewise 0.0 at 00Z Jan 1). Under the current 0-based convention, declination crosses zero at 00Z March 22 (non-leap); under a 1-based reading of the same constant it would be March 21.
  The real March equinox is ~March 20, so 1-based-with-80 would be marginally closer — but the formula's own approximations (fixed 23.5° obliquity, uniform 360/365°/day solar motion, hour-only day fraction) carry errors of the same one-to-two-day magnitude, so neither reading is "correct," and reinterpreting the base without retuning the 80 would silently shift declination by a 
  day everywhere.                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                             
  The related sub-day quirk is also fine physically: julian ignores minutes/seconds, but it only feeds declination (which drifts ≤ ~0.4°/day) and phenology (monthly resolution). The fast intra-day signal — the hour angle in cosz — is computed from TLOCTIM, which does include minutes and seconds. So hour-resolution julian is adequate for both of its consumers.                    
                                                                                                                                                                                                                                                                                                                                                                                             
  Bottom line                                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                             
  The code is internally coherent as a 0-based system: producer, both consumers, and the init value all agree, and the phenology consumer documents the convention. The lone 1-based statement — the module header — is simply wrong, and "fixing the code to match the docs" would be the worse direction: it would perturb both LAI interpolation and declination by a day for no physical 
  gain. The truth-restoring change is one line: correct the header comment to "day of year from 0 to <365/366> (0 at 00Z Jan 1)". If you do that, note the knock-on for the test suite: the "julian should be 1-based per docs" XFAIL row loses its rationale (its whole premise is the doc/code conflict) and should be dropped or converted, while the "julian should include minutes"     
  XFAIL row still stands on its own.

PhilMiller: So, given that, I'll change the comment and amend the test cases.

hellkite500 (approved):

JoshCu:
I was profiling some calibration just now and saw 24% of all cfe+nom execution was this! rebuilt from main and each calibration iteration has dropped from 3.8s to 2.9s :tada: 
great update!

PR #133, opened 2026-08-12
Build using CMake

robertbartel:
Refactor build setup to use CMake build system. Add targets for stand-alone driver (mainly for testing), BMI shared library, and test executable with existing, simple test routine.

# Additions
* New CMakeLists.txt file, with targets for standalone executable, shared library, and two test suites
* Flags for various compilers
* FindNetCDF routine for bring in that dependency

# Removals
* Makefiles

# Changes
* Changes to CI config to support CMake builds
* Changes to related parts of documentation files


cmake/CompilerFlags.cmake
```diff
@@ -0,0 +1,32 @@
+# Per-compiler, per-config Fortran flag matrix for noah-owp-modular.
+#
+# Applies Debug/Release flags via generator expressions so a single build tree
+# can switch configurations and multi-config generators pick the right set.
+
+function(noahowp_apply_fortran_flags target)
+    set(_id "${CMAKE_Fortran_COMPILER_ID}")
+
+    if(_id MATCHES "GNU")
```
PhilMiller, line 9: Could these be refactored a bit to have a general `_compiler_language_options` variable that contains the line length, preprocessor, math, etc options that are otherwise being duplicated across `_release` and `_debug`? 

And also fix `RelWithDebInfo` to actually be that.

CMakeLists.txt
```diff
@@ -0,0 +1,254 @@
...
+set(NOAHOWP_BMI_SOURCES
+    bmi/bmi.f90
+    bmi/bmi_noahowp.f90)
+
+set(NOAHOWP_DRIVER_SUPPORT_SOURCES
+    driver/AsciiReadModule.f90
+    driver/OutputModule.f90)
+
+set(NOAHOWP_DRIVER_SOURCES
+    ${NOAHOWP_DRIVER_SUPPORT_SOURCES}
```
PhilMiller, line 124: Could we skip the list append here, and just list `DRIVER_SUPPORT_SOURCES` among the dependencies for the executable?

PhilMiller (approved):
My comments are nice-to-haves, not requirements.

PR #135, opened 2026-08-15
First implementation of ngen BMI serialization protocol

robertbartel:
Implements the (pending) [ngen BMI serialization protocol](https://github.com/NOAA-OWP/ngen/pull/957).  Initial implementation geared primarily to supporting _hotstart_ scenarios that should not restore full state (i.e., _resume_ restores of full checkpoints).  Designed to prepare for future use of _resume_ restores as well, once mechanism is in place to communicate to model which is in use.

Builds on the earlier NGWPC serialization work, present in the NGWPC/development branch, while adapting it for aforementioned ngen protocol.  In particular, adheres to the same set of properties being serialized and deserialized as in the NGWPC version, excepted for deserialization of time and time step properties (left out of restore because it's wrong for _hotstart_).

Note that it is likely other state properties should also be omitted from _hotstart_ restore deserialization (and/or included within serialization/deserialization that aren't yet).  However, that will be scoped to a different issue/PR; this is just for the initial functional port of serialization functionality compatible with the ngen protocol, and so is using the same set of state properties as in the aforementioned NGWPC implementation this was based on.

> [!IMPORTANT]
> I'm actually almost certain that there are some fields the ported serialization/deserialization doesn't quite get right.  For now, I'm deliberately not fixing those, and carrying things over the way the NGWPC implementation handles them.  Confirmation and correction of those issues will be scoped separately.

## Additions

- Vendored [fortran-messagepack](https://github.com/synthfi/fortran-messagepack) library (including attribution in TERMS.md)
- Serialization/deserialization implementation using messagepack
- trivial logging module implementation, since NGWPC serialization changes were made after other logging-relate changes not yet merged to this repo
- two new test modules

## Changes

- Protocol-specific BMI variables in appropriate BMI functions for protocol, used to actually trigger serialization and deserialization
- Restore supporting different variations depending on restore mode, though just defaulting to _hotstart_  for now
- Update build config for vendored sources, new modules, and tests
- CI changes to include shared library build and new tests

## Notes

- The state variable is declared integer rather than a byte type, because the _iso_c_fortran_bmi_ middleware used to integrate with ngen does not support a `byte` type

## Todos

- Establish way of determining/communicating/receiving which restore method should apply, rather than just always using default
- A more thorough analysis and potential modifications to exactly which properties should be included in serialization/deserialization, and which should be restored in _hotstart_ versus _resume_ modes


extern/fortran-messagepack/README.md
```diff
@@ -0,0 +1,60 @@
...
+Third-party MessagePack serialization library for Fortran, used to encode and decode model state
+snapshots.
+
+| | |
+|---|---|
+| Upstream | https://github.com/synthfi/fortran-messagepack |
+| Commit | `2b2c9fc78bf3f7212ec9e35c675febdd508b7830` (2025-02-10) |
+| License | MIT, Copyright (c) 2025 Kelly Schultz — see [LICENSE](LICENSE) |
+
+## Where the sources are, and why
```
hellkite500, line 12: We should probably just update the ngen build to take advantage of the updated build system here anyways...

PhilMiller: Agreed - much preferable to reduce that obsolete complexity, and also reduce the necessary documentation here.

Especially since the CMake file in ngen would otherwise have to incorporate the quirks of building those files itself (the recursive function call flags).

extern/fortran-messagepack/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+```diff
+ ! messagepack_value.f90, mp_float_type
+-        logical :: is_64 = .true.
++        logical :: is_64 = .false.
+```
+
+Not carried here because it is inert. `is_64` is a default initializer reached only by a
+default-constructed `mp_float_type`; construction goes through the `mp_float_type` generic
+interface, whose `new_real32` and `new_real64` specifics both assign it explicitly.
```
PhilMiller, line 60: This belongs in a commit message or the PR description, not documentation of the code itself.

src/StateSerialization.f90
```diff
@@ -0,0 +1,985 @@
...
+        else if (index .EQ. 3) then
+            call get_int(mp_arr%values(index)%obj, deserialized_int, status)
+        else if (index .GE. 4) then
+            if (is_arr(mp_arr%values(index)%obj)) then
+                call get_arr_ref(mp_arr%values(index)%obj, mp_sub_arr, status)
+            end if
+        end if
+        select case(index)
+            case(1)
+                if (resume) domain%curr_datetime = deserialized_val
```
PhilMiller, line 232: It makes sense to me that `time_dbl` and `ITIME` are gated on `resume`.

Gating `curr_datetime` doesn't make sense to me, though. It corresponds to the physical state being restored, regardless of what the code is going to be asked to do after restoring the state. The case I can see for conditionalizing it is a distinct one, in which a saved state from simulated time A is used as the non-zero initialization for an ahistorical 'hotstart' at some distinct simulated time B (in lieu of a pure cold start + spinup).

src/StateSerialization.f90
```diff
@@ -0,0 +1,985 @@
...
+        end select
+    end do
+END SUBROUTINE domain_deserialization
+
+SUBROUTINE energy_serialization (energy, mp_arr)
+    type(energy_type), intent(in) :: energy
+    class(mp_arr_type), allocatable, intent(out) :: mp_arr
+
+    mp_arr = mp_arr_type(125)
+    mp_arr%values(1)%obj = mp_float_type(energy%cosz) !cosz
```
PhilMiller, line 250: The comments repeating the variable name touched by each statement seem pointless to preserve.

Maybe there'd be some value in formatting things so that the `=` all line up across indices, and break up different types with some extra newlines, so that it becomes easier to visually scan for differences from one line to the next.

src/StateSerialization.f90
```diff
@@ -0,0 +1,985 @@
...
+    mp_arr%values(91)%obj = mp_float_type(energy%SHB) !SHB
+    mp_arr%values(92)%obj = mp_float_type(energy%EVB) !EVB
+    mp_arr%values(93)%obj = mp_float_type(energy%GHB) !GHB
+    mp_arr%values(94)%obj = mp_float_type(energy%TAUXB) !TAUXB
+    mp_arr%values(95)%obj = mp_float_type(energy%TAUYB) !TAUYB
+    mp_arr%values(96)%obj = mp_float_type(energy%EHB2) !EHB2
+    mp_arr%values(97)%obj = mp_float_type(energy%T2MB) !T2MB
+    mp_arr%values(98)%obj = mp_float_type(energy%Q2B) !Q2B
+    mp_arr%values(99)%obj = mp_float_type(energy%QMELT) !QMELT
+    mp_arr%values(100)%obj = mp_float_type(energy%SNOWT_AVG) !SNOWT_AVG ,  could be realMissing
```
PhilMiller, line 349: This comment seems to want to say something meaningful, but I can't make sense of it.

src/StateSerialization.f90
```diff
@@ -0,0 +1,985 @@
...
+    mp_arr%values(101)%obj = mp_bool_type(energy%frozen_ground) !frozen_ground
+    mp_arr%values(102)%obj = mp_bool_type(energy%frozen_canopy) !frozen_canopy
```
PhilMiller, line 351:
```suggestion

    mp_arr%values(101)%obj = mp_bool_type(energy%frozen_ground) !frozen_ground
    mp_arr%values(102)%obj = mp_bool_type(energy%frozen_canopy) !frozen_canopy

```

src/StateSerialization.f90
```diff
@@ -0,0 +1,985 @@
...
+    mp_arr%values(103)%obj = transfer_values_to_mp(energy%FTDI) !FTDI array  (1:2)
```
PhilMiller, line 352: There's no point in commenting on the array bounds on all of these - the function being called serializes the entirety of each array by querying its metadata, regardless.

src/StateSerialization.f90
```diff
@@ -0,0 +1,985 @@
...
+    mp_arr%values(45)%obj = mp_float_type(water%SICEMAX) !SICEMAX
+    mp_arr%values(46)%obj = mp_float_type(water%FCRMAX) !FCRMAX
+    mp_arr%values(47)%obj = mp_float_type(water%FACC) !FACC
+    mp_arr%values(48)%obj = mp_float_type(water%QDRAIN) !QDRAIN
+    mp_arr%values(49)%obj = mp_float_type(water%DEEPRECH) !DEEPRECH
+    mp_arr%values(50)%obj = mp_float_type(water%ZWT) !ZWT
+    mp_arr%values(51)%obj = mp_float_type(water%ASAT) !ASAT
+    mp_arr%values(52)%obj = mp_float_type(water%SMCWTD) !SMCWTD
+    mp_arr%values(53)%obj = mp_int_type(int(water%ISNOW, kind=int64)) !ISNOW integer
+    mp_arr%values(54)%obj = transfer_values_to_mp(water%BTRANI ) !BTRANI array (1:levels%NSOIL)
```
PhilMiller, line 708:
```suggestion
    mp_arr%values(52)%obj = mp_float_type(water%SMCWTD) !SMCWTD

    mp_arr%values(53)%obj = mp_int_type(int(water%ISNOW, kind=int64)) !ISNOW integer

    mp_arr%values(54)%obj = transfer_values_to_mp(water%BTRANI ) !BTRANI array (1:levels%NSOIL)
```

src/StateSerialization.f90
```diff
@@ -0,0 +1,985 @@
...
+    type(mp_arr_type) :: dest
+    integer :: lb, ub, index, arr_size
+
+    lb = LBOUND(src,1)
+    ub = UBOUND(src,1)
+    arr_size = size(src)
+    dest = mp_arr_type(arr_size)
+    
+    do index = lb, ub
+        dest%values(index)%obj = mp_float_type(src(index))
```
PhilMiller, line 923: Accessing `dest%values(lb:ub)` seems erroneous in the cases where `lb` is anything other than `1`, as in all of the arrays above commented as using negative indices down to `NSNOW`.

I guess the comments I dismissed told me something after all.

I think this should be

```suggestion
        dest%values(index-lb+1)%obj = mp_float_type(src(index))
```

PhilMiller: The same concern applies to all of the `transfer_values_{to,from}_mp*` functions here

src/StateSerialization.f90
robertbartel: I see `curr_datetime` as an artifact of staging and executing a simulation, but not really a property that the simulation models.  Its value at any time step only depends on the initial configuration of the simulation.  As such, the value "belongs" to a particular simulation, and shouldn't really be used beyond that.  There may be fewer or smaller resulting problems from violating that for `curr_datetime` than for other things, but I think that's incidental, and you gave a tangible example of it not being completely safe.  

I will grant that for certain things, these lines get blurry, but I don't think there's any reason the meaning of `curr_datetime` needs to extend beyond a given simulation with respect to state saving and resuming.  But we can discuss more if there is something I'm not seeing.

src/StateSerialization.f90
robertbartel: You're not wrong, but this comes directly out of the NGWPC work.  I'd wanted to change as little of that as I could at this phase.  There is a risk this kind of thing gets totally forgotten about, but it also isn't causing any issues.  If you feel strongly, let me know and I'll change things, but I lean toward leaving this alone for now.

src/StateSerialization.f90
robertbartel: Same as my earlier (PR) comment on (code) comments:  this came straight out of the NGWPC work.  As such, I'd like to leave as is for now.  Again, if there are strong feelings otherwise, let me know.

src/StateSerialization.f90
robertbartel: Is this just for readability?  As with a few earlier things, if there isn't a problem with how this functions, nor a reason why this needs to change to support the protocol, I'd prefer to make as few changes as possible at this state with respect to the NGWPC code this was taken from.

src/StateSerialization.f90
robertbartel: Similar to other things, this came right from the inspiring NGWPC code.

src/StateSerialization.f90
robertbartel: Another things that appears to be just for readability and involves code from the NGWPC contributions.   Let me know if I'm just missing something subtle, or you feel very strongly, but otherwise I'd prefer to leave as-is.

src/StateSerialization.f90
PhilMiller: Ok, that's fine. These formatting/commenting remarks are all more in the vein of "how I'd like to see the code going forward", but easing comparison and merges with NGWPC work can take priority.

src/StateSerialization.f90
PhilMiller: On this one, I'd actually like to figure out what it's getting at. If it's something we need to pay attention to, then I think we should elaborate on this comment in particular. That's in spite of the NGWPC-divergence impact.

src/StateSerialization.f90
robertbartel: I think `lb` will always be `1` due to the way Fortran procedure array param assumed-shape and indices (re)numbering works.   So, from what I can tell (and to be fair my Fortran skills aren't the best), this is safe.

That's not to say it is good, but again, trying to carry over the NGWPC work with minimal changes, so I'd prefer to leave it for now.

robertbartel: FWIW, my above comment was only intended to applied to Phil's first one, as (despite it having been posted hours ago), my GitHub UI wasn't displaying Phil's "applies to all" follow-up.

The assumed-shape rule does apply to the other `*_to_*` procedure, but not the `*_from_*` ones.  I think they also will always have `lb`  as `1` for different reasons, but I'm confirming.

robertbartel: Well, actually, for the `*_from_*` procedures, I don't think it matters what `lb` is.  The `dest` array is being allocated with the same bounds as the loop.

src/StateSerialization.f90
robertbartel: It seems `realMissing` is a constant (well, maybe not a constant, but a local variable that is set and never changed) defined in _src/SnowWaterModule.f90_ as `-999999.0`.  Further, it appears, depending on conditions, `energy%SNOWT_AVG` might have been set to `realMissing`.   

So, in summary, this seems to be just making a note that this property might have been set to that placeholder value when a "real" value was missing.  Perhaps @ajkhattak or @seyounger can confirm that for us.

seyounger: @robertbartel Yeah, I think you are on the right track. The comment seems intended to point out that `energy%SNOWT_AVG` is set to `realMissing` when there is no snow present.

src/StateSerialization.f90
PhilMiller: I think the whole loop could be replaced by a vectorized statement

> `dest%values(:)%obj = mp_float_type(src(:))`

PhilMiller: Meanwhile, we've checked that `lbound/ubound` here are actually always returning 1 and N, respectively, and not the caller's bounds on the array, which sometimes range across (e.g.) `-NSNOW:NSOIL`

extern/fortran-messagepack/README.md
PhilMiller: Bobby and I discussed and concluded that we should leave this alone for the moment, and deal with it as part of a larger technical debt paydown effort. If we can get rid of the git submodules in ngen for the various BMI models, then this becomes moot, for instance.

